PR #115, opened 2015-02-13
Mel lain issues 1

MelLain:


src/artm/regularizer_interface.h
```diff
@@ -36,7 +37,8 @@ class RegularizerInterface {
                                const ModelConfig& model_config,
                                int inner_iter,
                                double tau,
-                               ::artm::utility::DenseMatrix<float>* theta) { return true; }
+                               ::artm::utility::DenseMatrix<float>* theta,
+                               std::shared_ptr<const artm::core::TopicModel> topic_model) { return true; }
```
sashafrey: std::shared_ptr<const artm::core::TopicModel>  => const TopicModel&
You API may cause some users to do this ugly thing:
http://stackoverflow.com/questions/12127758/pass-an-object-to-a-function-expecting-shared-ptr-without-actually-sharing-owner

src/artm_tests/instance_test.cc
```diff
@@ -197,8 +207,10 @@ TEST(Instance, MultipleStreamsAndModels) {
   test.instance()->CreateOrReconfigureModel(m2);
 
   for (int iter = 0; iter < 100; ++iter) {
-    test.instance()->local_data_loader()->InvokeIteration(1);
-    test.instance()->local_data_loader()->WaitIdle();
+    artm::InvokeIterationArgs inv_iter_args;
+    artm::WaitIdleArgs wait_args;
+    test.instance()->local_data_loader()->InvokeIteration(inv_iter_args);
+    test.instance()->local_data_loader()->WaitIdle(wait_args);
```
sashafrey: It is not important, but if you wish to same some lines there is a way to do so:
test.instance()->local_data_loader()->InvokeIteration(artm::InvokeIterationArgs());
test.instance()->local_data_loader()->WaitIdle(artm::WaitIdleArgs());

src/python/artm/library.py
```diff
@@ -487,7 +502,11 @@ def Overwrite(self, topic_model, commit = True):
 
     if commit:
       timeout = -1
-      HandleErrorCode(self.lib_, self.lib_.ArtmWaitIdle(self.master_id_, timeout))
+      args = messages_pb2.WaitIdleArgs()
+      args.timeout_milliseconds = timeout
+      args_blob = args.SerializeToString()
+      args_blob_p = ctypes.create_string_buffer(args_blob)
+      HandleErrorCode(self.lib_, self.lib_.ArtmWaitIdle(self.master_id_, len(args_blob), args_blob_p))
```
sashafrey: [Minor comment] Here it might be nice to reuse WaitIdle() method from the python interface -- without arguments. I should have done it earlier (when I've added the commit operation) -- in this case it would be nothing for you to do when ArtmWaitIdle changed its signature.

src/artm/core/batch_manager.cc
```diff
@@ -23,6 +23,23 @@ void BatchManager::DisposeModel(const ModelName& model_name) {
   in_progress_.erase(model_name);
 }
 
+void BatchManager::AddAndNext(const BatchManagerTask& task) {
+  boost::lock_guard<boost::mutex> guard(lock_);
+  tasks_.push_back(task);
```
sashafrey: >tasks_.push_back(task);
This line is redundant, please remove. Note that BatchManager::Next erases the task it returns from tasks_ list.

src/artm/core/batch_manager.h
```diff
@@ -46,6 +46,8 @@ class BatchManager : boost::noncopyable, public Notifiable {
   // until it is marked as processed by Done().
   BatchManagerTask Next();
 
+  void AddAndNext(const BatchManagerTask& task);
```
sashafrey: Overall I don't like method names in BatchManager -- Add, Next, and now AddAndNext makes very little sense to anyone (including me).
- Add() adds the a passive task to the queue
- Next() finds some passive task and moves it to the active queue (for each model)
- AddAndNext() puts the task directly to the active queue (for each model)
- Done() marks an active task as "done" (for a given model)
If you can find a better names for this operations, please rename. If not, just copy my test above as a comment into the code :)

src/artm/core/data_loader.cc
```diff
@@ -102,8 +102,11 @@ LocalDataLoader::~LocalDataLoader() {
   }
 }
 
-void LocalDataLoader::AddBatch(const Batch& batch, bool invoke) {
+bool LocalDataLoader::AddBatch(const AddBatchArgs& args, bool invoke) {
```
sashafrey: Could you try removing the 'bool invoke' argument?
I think you can get the same value from MasterComponentConfig.online_processing_mode

src/artm/core/data_loader.cc
```diff
@@ -342,6 +364,11 @@ void RemoteDataLoader::ThreadFunction() {
         break;
       }
 
+      if (instance()->schema()->config().online_batch_processing()) {
```
sashafrey: Please open a new issue to remove DataLoaderThread in online_processing_mode. However I think it might be really simple. The only thing you need to do is comment out this lines:

  // Keep this at the last action in constructor.
  // http://stackoverflow.com/questions/15751618/initialize-boost-thread-in-object-constructor
  boost::thread t(&LocalDataLoader::ThreadFunction, this);
  thread_.swap(t);

Then thread_ will refer to what boost calls 'not-a-thread'. Check http://www.boost.org/doc/libs/1_55_0/doc/html/thread/thread_management.html for the documentation.

src/artm/core/processor.cc
```diff
@@ -184,7 +185,8 @@ static void RegularizeAndNormalizeTheta(int inner_iter, const Batch& batch, cons
       continue;
     }
 
-    regularizer->RegularizeTheta(batch, model_config, inner_iter, tau, theta);
+    regularizer->RegularizeTheta(batch, model_config, inner_iter, tau, theta,
+                                 merger.GetLatestTopicModel(model_config.name()));
```
sashafrey: The code in processor should use merger.GetLatestTopicModel() function only ONCE for each (batch, model) pair, because the entire batch must be handled with a single Phi matrix.

Calling this method again might result in a different topic model used by regularizers comparing to the other code in the processor. This feels really bad (think of adding/removing tokens and topics... crazy things might happen to regularizer in this case).

I think for now you should revert this change since it conflicts with my re-work in Processor.cc. Lets re-introduce it after my pull request is in the master.

src/artm/core/data_loader.cc
```diff
@@ -114,18 +117,36 @@ void LocalDataLoader::AddBatch(const Batch& batch, bool invoke) {
...
+      }
+    }
+    auto pi = std::make_shared<ProcessorInput>();
+    pi->mutable_batch()->CopyFrom(*modified_batch);
+    boost::uuids::uuid uuid = boost::uuids::random_generator()();
+    pi->set_batch_uuid(boost::lexical_cast<std::string>(uuid));
+    instance_->batch_manager()->AddAndNext(BatchManagerTask(uuid, std::string()));
+    instance_->processor_queue()->push(pi);
+  } else {
+    generation_->AddBatch(modified_batch);
```
sashafrey: Please, open a new issue 'support cache_theta flag in the online_processing_mode'. This is need in order to remove AddBatch() for non-online_processing_mode. As the result we will be able to clean-up things like MemoryGeneration and some other code.

PR #129, opened 2015-02-18
introduce reset_scores flag in InvokeIterationArgs and AddBatch args, cl...

MelLain:
...eanup tests, add 'clean_cache' flag

### DIFF
--- a/src/artm/core/data_loader.cc
+++ b/src/artm/core/data_loader.cc
@@ -82,10 +82,12 @@ LocalDataLoader::LocalDataLoader(Instance* instance)
     generation_.reset(new DiskGeneration(disk_path));
   }
 
-  // Keep this at the last action in constructor.
-  // http://stackoverflow.com/questions/15751618/initialize-boost-thread-in-object-constructor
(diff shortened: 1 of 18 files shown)


src/artm/core/data_loader.cc
```diff
@@ -82,10 +82,12 @@ LocalDataLoader::LocalDataLoader(Instance* instance)
     generation_.reset(new DiskGeneration(disk_path));
```
sashafrey: Remove MemoryGeneration class - I think it is no longer in use

src/artm/core/data_loader.cc
```diff
@@ -221,6 +221,10 @@ bool LocalDataLoader::RequestThetaMatrix(const GetThetaMatrixArgs& get_theta_arg
     } else {
       BatchHelpers::PopulateThetaMatrixFromCacheEntry(*cache, get_theta_args, theta_matrix);
     }
+
+    if (get_theta_args.clean_cache()) {
+      cache.reset();
```
sashafrey: cache.reset() -> cache_.erase(key)

PR #150, opened 2015-03-08
perplexity was modified to work with modalities

MelLain:


bigartm:
Fixed by https://github.com/bigartm/bigartm/pull/151 (backward-compatible change in messages.proto)

PR #152, opened 2015-03-08
LabelRegularization regularizer was added.

MelLain:


bigartm:
This is the same request as https://github.com/bigartm/bigartm/pull/191

PR #173, opened 2015-04-02
OnlineLDA benchmark

romovpa:


bigartm:
Thanks for a super benchmark! I'll try to run it myself somewhere within next week, and then integrate to Master branch.

If you have time, it would be also great to write a short page in documentation describing this experiment (similar to 
http://docs.bigartm.org/en/latest/stories/experiment02_artm.html that describes https://github.com/bigartm/bigartm/blob/master/src/python/experiments/experiment02_artm.py).

sashafrey:
This should be rewritten as a notebook, correct? I still had no time to run this experiment myself, however if we turn it into notebook then I'll definitely learn how to run it.

romovpa:
>This should be rewritten as a notebook, correct? 

This benchmark runs for several hours on several machines and does neither interactive nor visual. Rewriting it as a notebook is meaningless. We just need a complete readme.

sashafrey:
Ok, agree.. Then I put it this into "ready for review". You are done, and the ball is on my side - I just need to review the code and learn how to run this benchmark. So far I had no time to do this, but I'll do my best to not delay it any further. At the same time I will really appreciate if someone else has time to verify this benchmark; then we will merge it in.

sashafrey:
I rebase this to ``bigartm/master`` and merged in https://github.com/bigartm/bigartm/pull/295. This experiment is based on the old API, but it is still useful to have it in master if we would like to reproduce this results on latest bigartm.

I'm sorry to keep this as a pull request for so long.

PR #181, opened 2015-04-08
[python] Verbose error handling on missing library

IIoTeP9HuY:
I'm getting 
 "dlopen(libartm.dylib, 6): image not found, fall back to ARTM_SHARED_LIBRARY environment variable"
error message in stdin each time I create artm.library.Library() object. However, this error is not critical because of the fallback. I believe this kind of messages should be logged with lower log level, for example to stderr.

### DIFF
--- a/src/python/artm/library.py
+++ b/src/python/artm/library.py
@@ -100,6 +100,8 @@ class Library:
         if not artm_shared_library:
             if sys.platform.count('linux') == 1:
                 artm_shared_library = 'libartm.so'
+            elif sys.platform.count('darwin') == 1:
+                artm_shared_library = 'libartm.dylib'
             else:
                 artm_shared_library = 'artm.dll'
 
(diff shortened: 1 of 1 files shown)


sashafrey:
Thanks! Sure, you are absolutely right -- let's fix this.

I've been thinking what should be the overal logic here. The cleanest way is perhaps as follows:
1. If ARTM_SHARED_LIBRARY env variable exists, try use its value as a full path to artm shared library. If that failed - return an error.
2. Fall back to find_library("artm") from ctypes (https://docs.python.org/2/library/ctypes.html)

The artm_shared_library argument of __init__ method is useless in my opinion, and should be simply removed. If somebody would like to define path to the library in the code it would be more straighforward to set ARTM_SHARED_LIBRARY env variable from python:
``os.environ['ARTM_SHARED_LIBRARY'] = ... ``. 
Removing it is not a problem - I can mention it in release notes for next BigARTM version and I don't think it will cause big inconvenience to anyone.

I can just accept your request because it is a step further, and put my comments into a separate so we can address them later. Let me know what do you think. 

Thanks again!
Alex

IIoTeP9HuY:
Hi!
I feel like find_library(...) is the right way to go. In similar cases people tend to set LD_LIBRARY_PATH and DYLD_LIBRARY_PATH variables to use libraries in non-default paths, and find_library should handle it correctly.
Yet, variable ARTM_SHARED_LIBRARY variable may probably be helpful during development.

Parameter "artm_shared_library" can be helpful in case you want to unit test library loading and don't want several tests that run in parallel to clash on environment variables (they are shared between process and it's children). Though, I'm not sure whether this test is worth keeping this parameter :)

I agree that we should move iteratively, and this change doesn't break any compatibility.

sashafrey:
Agree!

PR #183, opened 2015-04-09
[python] Fix segfault during conversion between c and python string

IIoTeP9HuY:
On OSX, on any internal error (missing parsed file, etc.) program exits with segfault without any error message.
This problem is described here: http://stackoverflow.com/questions/14944367/ctypes-segfault-on-osx-only
Not sure, whether it's ctypes bug, but this diff fixes problem on my OSX and works correctly on Ubuntu 14.04 VM.

### DIFF
--- a/src/python/artm/library.py
+++ b/src/python/artm/library.py
@@ -66,8 +66,8 @@ class NetworkException(BaseException): pass
 
 
 def GetLastErrorMessage(lib):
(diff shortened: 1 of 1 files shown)


sashafrey:
Thanks! And as far as I remember this is the only place that we need to change.  Elsewhere in artm.library we allocate memory in ctypes, and ask BigARTM to fill it with data.

PR #186, opened 2015-04-09
MelLain pull request

MelLain:


src/artm/core/protobuf_helpers.h
```diff
@@ -49,6 +50,38 @@ void repeated_field_append(T* field, int index, V value) {
   field->Set(index, new_value);
 }
 
+template<class T>
+std::vector<bool> is_member(T a, T b) {
```
sashafrey: - Pass arguments by reference to avoid unnecessary  copying 
- Rename argument names: "a" to "elements" and "b" to "set" (I think matlab's names "a" and "b" are terribly bad - I always need to go to documentation to figure out what order should I use). You may also add a comment that "is_member finds which elements are present in the set. The resulting vector has the same size as elements size".

template<class T>
std::vector<bool> is_member(const T&elements, const T&set)

src/artm/core/protobuf_helpers.h
```diff
@@ -49,6 +50,38 @@ void repeated_field_append(T* field, int index, V value) {
...
+  std::vector<bool> retval;
+
+  if (a.size() > 0) {
+    for (int i = 0; i < b.size(); ++i)
```
sashafrey: From what I see you verify whether elements of "b" are present in "a" array. Matlab version works the other way around. I suggest to replace the order of you arguments to be consistent with Matlab's ismember.

src/artm/core/protobuf_helpers.h
```diff
@@ -49,6 +50,38 @@ void repeated_field_append(T* field, int index, V value) {
...
+
+    for (int i = 0; i < a.size(); ++i)
+      for (int j = 0; j < b.size(); ++j)
+        if (b.Get(j) == a.Get(i)) {
+          retval[j] = true;
+          break;
+        }
+  } else {
+    for (int i = 0; i < b.size(); ++i)
+      retval.push_back(true);
```
sashafrey: use ''false'' by default? If target set is empty - then ismember=false.

src/artm/regularizer_sandbox/decorrelator_phi.cc
```diff
@@ -17,31 +18,11 @@ bool DecorrelatorPhi::RegularizePhi(::artm::core::Regularizable* topic_model, do
...
-          topics_to_regularize[real_topic_id] = true;
-          break;
-        }
-      }
-    }
-  } else {
-    for (int i = 0; i < topic_size; ++i)
-      topics_to_regularize.push_back(true);
-  }
+  std::vector<bool> topics_to_regularize = core::is_member(config_.topic_name(),
```
sashafrey: Now I see why is_member returns "true" by default... I think this is still beter to return "false" by default, because this is consistent with matlab.

std::vector<bool> topics_to_regularize;
if (config_.topic_name().empty())
   topics_to_regularize.assign(config_.topic_name().size(), true)
else
  topics_to_regularize = core::is_meber(config_.topic_name());

This is more verbose, but then you explicitly see how we treat empty topic_name.

sashafrey: Or just introduce "use_all_topics" flag and check "if (use_all_topics || topics_to_regularize[topic_id])". This is safe because C++ guaranties to NOT EVALUATE second part of "if (A || B)" whenever A evaluates to true.

http://stackoverflow.com/questions/1331318/how-an-if-a-b-statement-is-evaluated
In C and C++, the && and || operators "short-circuit". That means that they only evaluate a parameter if required. If the first parameter to && is false, or the first to || is true, the rest will not be evaluated.

src/artm/regularizer_sandbox/label_regularization_phi.cc
```diff
@@ -0,0 +1,104 @@
...
+      }
+    }
+  }
+
+  return true;
+}
+
+bool LabelRegularizationPhi::Reconfigure(const RegularizerConfig& config) {
+  std::string config_blob = config.config();
+  SmoothSparsePhiConfig regularizer_config;
```
sashafrey: SmoothSparsePhiConfig ??

src/artm/regularizer_sandbox/label_regularization_phi.cc
```diff
@@ -0,0 +1,104 @@
...
+  for (int token_id = 0; token_id < token_size; ++token_id) {
+    auto token = topic_model->token(token_id);
+
+    if (use_all_classes || core::is_member(token.class_id, config_.class_id())) {
+      // count sum of weights
+      float weights_sum = 0.0f;
+      for (int topic_id = 0; topic_id < topic_size; ++topic_id) {
+        if (topics_to_regularize[topic_id]) {
+          // token_class_id is anyway presented in n_t
+          weights_sum += p_wt[token_id][topic_id] * n_t.find(token.class_id)->second[topic_id];
```
sashafrey: Ups, you lookup n_t.find(token.class_id) INSIDE topic_id loop. This is slow. Move it outside of the loop and assign like this:

auto class_iter = n_t.find(token.class_id);
assert(iter != nullptr)

Now it is safe to use iter plus there is one lookup in the map for every token.

src/artm/regularizer_sandbox/smooth_sparse_phi.cc
```diff
@@ -53,55 +37,36 @@ bool SmoothSparsePhi::RegularizePhi(::artm::core::Regularizable* topic_model, do
...
-            regularize_this_token = true;
-          }
-          if (regularize_this_token) {
-            if (dictionary_ptr->find(token) != dictionary_ptr->end()) {
-              float value = dictionary_ptr->find(token)->second.value() * static_cast<float>(tau);
-              topic_model->IncreaseRegularizerWeight(token_id, topic_id, value);
-            }
-          }
+        if (use_all_classes || core::is_member(token.class_id, config_.class_id())) {
+          float value = static_cast<float>(tau) * coeffs.find(token)->second;
```
sashafrey: Performance: avoid coeffs.find(token) lookup inside topic_id loop

src/artm/regularizer_sandbox/label_regularization_phi.cc
```diff
@@ -0,0 +1,104 @@
...
+        }
+      }
+      // form the value
+      for (int topic_id = 0; topic_id < topic_size; ++topic_id) {
+        if (topics_to_regularize[topic_id]) {
+          float p_c = coeffs.find(token)->second;
```
sashafrey: Crash when "has_dictionary==true" and token is missing from the dictionary. This really can happen because we have no control the dictionary - it is created by the user and he may not add some tokens there.

I don't see why you need "coeffs" map. I think it is more straightforward to find the coeff inside "for (int token_id = 0; ... " loop (after "//proceed the regularization" comment).

src/artm/regularizer_sandbox/smooth_sparse_phi.cc
```diff
@@ -53,55 +37,36 @@ bool SmoothSparsePhi::RegularizePhi(::artm::core::Regularizable* topic_model, do
     has_dictionary = false;
   }
 
+  std::map<core::Token, float> coeffs;
```
sashafrey: Same questions as for label_regularizer.cc:
- avoid coeffs variable and find the coeff "on fly"?
- handle gracefully user dictionary with missing tokens
- move expensive stuff out from the inner loop:
--- core::is_member(token.class_id, config_.class_id()
--- coeffs.find(token)

src/artm/core/protobuf_helpers.h
```diff
@@ -49,6 +50,33 @@ void repeated_field_append(T* field, int index, V value) {
   field->Set(index, new_value);
 }
 
+template<class T>
+std::vector<bool> is_member(const T& elements, const T& set) {
+  std::vector<bool> retval;
+  retval.assign(set.size(), false);
```
sashafrey: Matlab version has
retval.assign(elements.size(), false);

src/artm/regularizer_sandbox/label_regularization_phi.cc
```diff
@@ -0,0 +1,103 @@
...
+  // proceed the regularization
+  for (int token_id = 0; token_id < token_size; ++token_id) {
+    auto token = topic_model->token(token_id);
+    auto class_iter = n_t.find(token.class_id);
+    assert(class_iter != n_t.end());
+
+    float coefficient = 1.0f;
+    if (has_dictionary) {
+      if (use_all_classes ||
+          core::is_member(token.class_id, config_.class_id())) {
```
sashafrey: if (!use_all_classes && !core_is_member(...)) continue;
This just just an alternative - your code is also OK. I just don't like extra tabulation)

src/artm/regularizer_sandbox/smooth_sparse_phi.cc
```diff
@@ -53,55 +38,32 @@ bool SmoothSparsePhi::RegularizePhi(::artm::core::Regularizable* topic_model, do
...
+        // don't process tokens without value in the dictionary
+        if (entry_iter == dictionary_ptr->end())
+          coefficient = 0.0f;
+        else
+          coefficient = entry_iter->second.value();
       }
     }
-  } else {
-    // proceed the regularization
+    bool use_this_token = core::is_member(token.class_id, config_.class_id());
```
sashafrey: if (!use_all_classes && !use_this_token) continue;

In this case its nice to avoid "if" inside innermost loop. The code is equally readable, and second version is obviously more efficient. Also move "float value = ..." outside of the innermost loop since it is the same for all topics.

bigartm:
This is the same pull request as https://github.com/bigartm/bigartm/pull/191

PR #198, opened 2015-05-02
Relative reg coefs for Phi were added

MelLain:
Pre-review version with correct commit history


MelLain:
Was finished in another pull request

PR #203, opened 2015-05-15
Mel lain spec sparse reg 2 (merged with bigartm/master)

MelLain:


bigartm:
Merged in https://github.com/bigartm/bigartm/pull/204

PR #205, opened 2015-05-16
Mel lain minor

MelLain:
[] -> get()/set() for further usage in sparse p_wt mode


sashafrey:
Re-open request as https://github.com/bigartm/bigartm/pull/206

PR #213, opened 2015-05-26
MelLain Coherency

MelLain:
- coherency regularizer and scores were added
- dictionaries representation was modified
- parser was modified to store cooc info, tests were updated
- '..._sandbox' -> '...' for scores and regularizers
- python examples for coherency were added


bigartm:
Merged in https://github.com/bigartm/bigartm/pull/216

PR #221, opened 2015-06-05
Mel lain new python api (in separated branch)

MelLain:


MelLain:
Will be re-opened in new request soon.

PR #226, opened 2015-06-20
Mel Lain - New Python API

MelLain:


romovpa:
**[PEP-8](https://www.python.org/dev/peps/pep-0008/)**!!! >:/

PR #228, opened 2015-06-26
Mel lain new python api release

MelLain:


src/python/artm/_js/ldavis.js
```diff
@@ -0,0 +1,1361 @@
+LDAvis = function(to_select, data) {
```
sashafrey: What is the license for ldavis? Please check that it follows a permissive license (MIT, BSD-3-clause, or similar). Consider adding ldavis license at the bottom of BigARTM LICENSE file (in the root of the project)

.gitignore
```diff
@@ -78,6 +78,7 @@ src/Win32/*
 /3rdparty/glog/vsprojects/libglog_static/Release/*
 /3rdparty/rpcz/vsprojects/Intermediate/*
 /3rdparty/rpcz/vsprojects/Win32/*
+/3rdparty/protobuf/python/*
```
sashafrey: It is scary to ignore all files under 3rdparty/protobuf/python, some of them are NOT auto-generated. It will be a big issue if we decide to upgrade to the next protobuf version. A better way of ignoring auto-generated protobuf files would be to list them explicitly (or to list all auto-generated folders).

src/python/artm/visualization_ldavis.py
```diff
@@ -0,0 +1,125 @@
+import json
```
sashafrey: @romovpa Do you have any comments about this file?

src/python/vsprojects/python_examples.pyproj
```diff
@@ -5,7 +5,7 @@
     <SchemaVersion>2.0</SchemaVersion>
     <ProjectGuid>8cbe34f8-5d46-4832-b615-db183812f81f</ProjectGuid>
     <ProjectHome>.</ProjectHome>
-    <StartupFile>..\examples\example16_coherence_score.py</StartupFile>
+    <StartupFile>..\examples\example_artm_model.py</StartupFile>
```
sashafrey: I can't see example_artm_model.py in this pull request. Have you added it to git?

sashafrey:
artm_model.py introduces several new python-dependencies, that are not included in out-of-the-box python installation. For example, numpy. Please, test on a clean machine what are those dependencies, and list them in the tutorial section of the documentation

sashafrey:
I'm not sure if we are ready to expose gather_cooc and cooc_tokens in  ArtmModel.Parse(). If you need them for experiments I think we should mention in comments that this options are truly experimental and might be removed (or changed) in the very next release.

sashafrey:
num_processors - is int, default=1
Please revisit this as I made a new feature to auto-detect the number of CPU threads (according to http://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency). I think this is by far better than defaulting to 1 processor.

https://github.com/sashafrey/bigartm/commit/c8deda5b162c4d9cd67be4e4acbae119f38e2850
(pending pull request)

sashafrey:
elif data_format == 'bow_vw': raise NotImplementedError()
It would be great to have this option :)
The CollectionParserConfig is very similar to  bow_uci. The only difference is that you vocab file is not used. The real input is passed via docword file.

sashafrey:
To my taste it sounds better if we remove "_count" in parameters 
document_passes_count
collection_passes_count

sashafrey:
Consider another default options in fit_offline:
apply_weight=1.0
decay_wegith=0.0
This better reflects offline algorithm.

sashafrey:
DataFrame looks really cool.

values = [[item_w for item_w in item_weights.value] for item_weights in theta_matrix.item_weights]
retval = DataFrame(data=np.matrix(values).transpose(), columns=document_ids,index=topic_names)

src/python/vsprojects/python_examples.pyproj
MelLain: I've removed this file and it's line python_examples.pyproj in further commit.

src/python/artm/_js/ldavis.js
romovpa: There is no standard license: https://github.com/cpsievert/LDAvis/blob/master/LICENSE. Just a link to the copyright holder. :( 
I will ask the author.
Anyway, this code is just proof-of-concept and an impression for us. If there are problems with LDAvis permissions, we could easily make our original d3-based visualization.

src/python/artm/visualization_ldavis.py
romovpa: I wrote it :)

src/python/vsprojects/python_examples.pyproj
romovpa: @MelLain Sorry, is there an example of using new API in this pull request?

romovpa:
> artm_model.py introduces several new python-dependencies, that are not included in out-of-the-box python installation. For example, numpy. Please, test on a clean machine what are those dependencies, and list them in the tutorial section of the documentation
The easiest way is to add `requirements.txt` with a list of python dependencies and run `sudo pip install -r  requirements.txt` during installation. Also numpy and its family requires several system packages to be installed (like gfortran, blas, lapack).

sashafrey:
It would be convenient to add a getter for 

.value[model.num_phi_updates-1]

(e.g. the last calculated value of a score). I think this will be a very common scenario, and we should not ask user to type "model.num_phi_updates-1" every time he wants score value...

MelLain:
I have one question about this note: do we need such getters only for important fields of scores, like 'value' for perplexity, or for all fields? For instance, peprlexity score also has 'zero_tokens' and 'total_tokens' fields, do we need such getters for them?

Regards, Murat.

05.07.2015, 16:23, "sashafrey" <notifications@github.com>:

sashafrey:
I'm not sure -- of course its better to have it for all fields, but it means more code in ArtmModel and more importantly - harder to maintain ArtmModel as we add more features.

Have you considered a different solution where PerplexityScoreInfo (and other ScoreInfo-s) are scalar (e.g. they 1:1 correspond to protobuf messages), and then you keep track of historical values in ArtmModel._score_info. What I mean is to add a second key to score_info - the first is for score name (as you have now), and the second key is for num_phi_updates. Then all XXXScoreInfo classes became simpler because they don't need to track history. The only issue is that it may be more difficult for a user to access the list of historical values, because it will require iterating along "i" in

score_info[name, i].value()

Is there a pythonic way of doing this? In C# I'd simply write a LINQ select, so the API would be as convenient as if you make "value" itself a vector.

MelLain:
The main reason of such organization is to give user an opputunity to plot such scores, as perplexity, without need to access each element, e.g. in described way:

plot(model.score_info['peprlexity'].value)

score_info[name, i].value() is not such convenient.

Regards, Murat.

05.07.2015, 20:37, "sashafrey" <notifications@github.com>:

sashafrey:
Yep, that makes sense. I'll looked once again on XXXScoreInfo classes, and they are not too difficult to maintain. Let's keep them the way they are implemented now. Thanks for adding "_last" getters!

sashafrey:
Consider rename
topics_count => num_topics
tokens_count => num_tokens
(looks more consistent with the rest of the namings)

sashafrey:
Consider changing default num_document_passes to 10. This is more consistent with VW and Gensim, which use default 50 (but in addition have a threshold for early stop -- if p(t|d) changes are less than 1e-5 they they exit earlier).

sashafrey:
ArtmModel.parse
default data_format='batches'
According to documentation default is 'bow_uci', and 'batches' is not a valid input

sashafrey:
print 'No collection name was given, skip model.fit_offline()'
looks wrong inside get_theta()

romovpa:
Sorry for delay, I will review soon

sashafrey:
New API is really AWESOME. I do love it :)

sashafrey:
Since I tried it, here are some comments:
=============================
>>> model.parse(data_path='d:\datasets', collection_name='kos')
Unknown data format, skip model.parse()
[Wrong default?]

>>> model.parse(data_path='d:\datasets', data_format='bow_uci', collection_name='kos')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "artm\artm_model.py", line 2726, in parse
    unique_tokens = library.Library().ParseCollection(collection_parser_config)
  File "artm\library.py", line 153, in ParseCollection
    len(config_blob), config_blob_p))
  File "artm\library.py", line 97, in HandleErrorCode
    raise DiskReadException(GetLastErrorMessage(lib))
artm.library.DiskReadException: DiskReadException :  File d:\datasetsvocab.kos.txt does not exist.
[Need a better concatenation between path and filename?]



// in the notebook, first example
model = ArtmModel(num_processors=2, num_topics=15, num_document_passes=1)
[I think its best to remove num_processors and num_document_passes from the example; change it to "model = ArtmModel(num_topics=15)"]

// in the notebook, first example:
>>>model.initialize(dictionary=model.load_dictionary('kos/dictionary'))
[Please add a comment that this step is optional, and it will be automaticaly executed in fit_offline and fit_online when model is uninitialized. ]

>>> model.scores.add(TopTokensScore(name='TopTokensScore', num_tokens=5))
>>> for topic_name in model.topic_names:
...     print topic_name + ': ',
...     print model.scores_info['TopTokensScore'].last_topic_info[topic_name].tokens
...
@topic_0:
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
KeyError: 'TopTokensScore'

[That's interesting. I understand why this is happening --- user must run at least one iteration to get scores. That's OK for Theta scores, but technically its possible to retrieve Phi scores without running additional "fit". Let's think how should we support this scenario. I think this should be done via model.master. The ArtmModel API is really good, because it unifies phi- and theta- scores, and the user doesn't need to know the difference between them. If the user happens to know that some scores (e.g. phi scores) can be retrieved, then he needs to go to master. However master API doesn't give this info -- it require user to go via artm.Score. How do we solve this?]

>>>model.to_csv(file_name='kos_artm_model.csv')
Token;Class ID;TOPIC: @topic_0;TOPIC: @topic_1;TOPIC: @topic_2;TOPIC: @topic_3;TOPIC: @topic_4;TOPIC: @topic_5;TOPIC: @topic_6;TOPIC: @topic_7;TOPIC: @topic_8;TOPIC: @topic_9;TOPIC: @topic_10;TOPIC: @topic_11;TOPIC: @topic_12;TOPIC: @topic_13;TOPIC: @topic_14
aarp;@default_class;3.6410899794357687e-13;3.609851566971356e-12;3.910421621656623e-15;1.0261647958031972e-06;1.836142349276315e-11;1.7397466800517236e-09;1.716532610496202e-10;9.937288215607423e-14;2.6023754332027416e-11;1.9118295949182088e-15;1.0968549402079475e-09;0.00031675511854700744;1.1415996999716155e-13;1.5530036989730434e-06;3.541752812452614e-05
abandon;@default_class;1.2040577530569863e-05;1.9726703612832353e-06;0.00013656489318236709;7.230268965940922e-05;2.9698483103857143e-06;1.9598221570049645e-06;8.465028076898307e-05;5.340566322509233e-10;6.823032890679315e-06;2.538816443120595e-05;3.2243413443211466e-05;5.752855577156879e-05;0.00019831059034913778;0.0002477771486155689;1.2875776519649662e-05
a
....

[Should we add some options to make this more human-readable? E.g. round some digits? BigARTM users float precision, and what you output looks like double precision. Should we limit to up to 5 digits by default?]
=============================

But again, ArtmModel is awesome. Lets ship it ASAP :)

src/python/artm/_js/ldavis.js
romovpa: The license is MIT
Noted [here](https://github.com/cpsievert/LDAvis/blob/master/DESCRIPTION)

src/python/examples/example18_merge_model.py
```diff
@@ -36,7 +36,7 @@
             batches_to_process.append(batch_filename)
             if ((batch_index + 1) % update_every == 0) or ((batch_index + 1) == len(batches)):
                 master.ProcessBatches(pwt_model, batches_to_process, "nwt_hat")
-                master.MergeModel({("nwt", 0.7), ("nwt_hat", 0.3)}, target_nwt="nwt")
+                master.MergeModel({"nwt" : 0.7, "nwt_hat" : 0.3}, target_nwt="nwt")
```
romovpa: [pep8: avoid extraneous whitespace](https://www.python.org/dev/peps/pep-0008/#id20)

```python
Yes: spam(ham[1], {eggs: 2})
No:  spam( ham[ 1 ], { eggs: 2 } )
```

src/python/examples/example19_regularize_model.py
```diff
@@ -44,8 +44,8 @@
     phi_background = master.CreateSmoothSparsePhiRegularizer(topic_names=background_topics)
     decorrelator = master.CreateDecorrelatorPhiRegularizer(topic_names=objective_topics)
 
-    theta_regularizers = {(theta_objective.name(), -0.5), (theta_background.name(), 0.5)}
-    phi_regularizers = {(phi_objective.name(), -0.5), (phi_background.name(), 0.5), (decorrelator.name(), 100000)}
+    theta_regularizers = {theta_objective.name() : -0.5, theta_background.name() : 0.5}
+    phi_regularizers = {phi_objective.name() : -0.5, phi_background.name() : 0.5, decorrelator.name() : 100000}
```
romovpa: also here (extraneous whitespaces)

sashafrey:
Done :+1:

PR #236, opened 2015-07-06
add ability to load cooc info into parser

MelLain:


src/artm/core/dictionary.h
```diff
@@ -46,7 +46,7 @@ class Dictionary {
   std::vector<DictionaryEntry> entries_;
   std::unordered_map<Token, int, TokenHasher> token_index_;
   std::unordered_map<int, Token> index_token_;
-  std::unordered_map<int, std::unordered_map<int, int> > cooc_values_;
+  std::unordered_map<int, std::unordered_map<int, std::vector<float> > > cooc_values_;
```
sashafrey: I'm a bit concerned about storing so many std::vector<float>, particularly when each of them has only 1 value. Look at sizeof(std::vector<float>), which is 24 bytes on 64 bit. Plus the actual memory for 1 float. In total this is 7 times more memory usage than you really need. 

Again, I'm not convinced that we need to support multiple columns in cooc file - this makes the API more complicated, plus user needs to remember which columns is for what purpose. To me it is far more simple to create several dictionaries in BigARTM.  What do you think? 
If you really want to support multiple columns in cooc file then I would suggest to store them as follows:
std::vector<std::unordered_map<int, std::unordered_map<int, int> >>  cooc_values_.
I think this will use less memory (even though it duplicates all keys).
This option will only work if all of your std::vector<float> had the same length, but I think this is required anyway...

MelLain: The main reason against several dictionaries is that we need to run parser every time we want to create new dictionary. If the parser provides option to create dictionary using cooc and vocab files, without looking through docword, it will be better to implement one value.

Regards, Murat.

06.07.2015, 20:42, "sashafrey" <notifications@github.com>:

sashafrey: True, I haven't thought about this. Then how about this solution:
1. Add a new format into CollectionParserConfig.Format enum (I vote for calling it "Cooccurrence", but any reasonable name will do)
2. Treat 'docword_file_path' as concurrence file (e.g. with 3 columns)
3. Output dictionary file with coocurrences parsed from docword_file_path
4. Do NOT produce any batches

sashafrey: What I really like about this solution is that it only adds one line to messages.proto (the line that adds new format to enum). That's all. Also, the user will have to call "ArtmParseCollection" multiple times - once to produce the batches, and then (potentially multiple calls) to produce dictionaries with cooc information. To me this sounds reasonable, because making batches and making dictionaries are independent operations, so I'm very happy to keep this as separate as possible.

MelLain: I think, we have one problem. Coherence counting require information about total items count in collection and items count for each entry token. It's impossible to retrieve this values without docword:(

06.07.2015, 22:41, "sashafrey" <notifications@github.com>:

sashafrey: Not if we say that cooc file contains PMI. Then you simply count coherence as described in https://s3-eu-west-1.amazonaws.com/artm/voron15slds.pdf (page 7, last paragraph). 

coh = 2/(k * (k-1)) sum_{i, j} PMI(w_i, w_j)               (1)

The value (last column) in cooc file needs to be PMI(w_i, w_j), and BigARTM do no further processing of this number. However, BigARTM does not care if the user have actually calculated PMI or not -- we just calculate via a simple formula (1).

Does this solve the issue, or I'm still missing something important?

MelLain: Well, different formulas of coherence require different values, including collection characteristics. It seems to be incorrect to limit user with only described formula. Anya and I decided to implement several modes in coherency plugin with different formulas. So I think, it's necessary to collect such values, as total_items_count and entry.items_count

Maybe we can ask for dictionary file, collected during parsing, as a third parameter of parsing in cooc mode and add necessary data from file into it. The only limit is that user must guarantee that numbers of tokens in dictionary and vocab file should be equal and tokens must have the same order. Is this way too complex?

Regards, Murat.

07.07.2015, 00:31, "sashafrey" <notifications@github.com>:

sashafrey: Ok, I thought that
- all types of coherence fit simple formula (1), e.g. calculate an average across of VALUE(w_i, w_j) for all pairs in a specific list (like top tokens, or topic kernels). 
- and all complaxity is hidden in a way VALUE(w_i, w_j) is calculated. So the plan was to push this complexity to the user, and keep BigARTM clean :)

MelLain: Okey, I give in:)

07.07.2015, 00:46, "sashafrey" <notifications@github.com>:

sashafrey: Do you have that discussion somewhere in e-mail? I would like to see examples of what does not fit (1)

MelLain: No, not any e-mail, we discussed it during meeting. It'd be better to ask Anya about it.

07.07.2015, 00:48, "sashafrey" <notifications@github.com>:

PR #237, opened 2015-07-06
ArtmCopyRequestResultEx for explicit phi/theta retrieval

sashafrey:
This is an initial commit that begins the implementation of explicit phi and theta matrices retrieval. For now it only implements the retrieval for C++. The next change is to consume this new API from Python, and produce a Python matrix (or to produce pandas.dataframe). The rationale behind this change is to reduce the size of TopicModel and ThetaMatrix messages, produce by c_interface, and to reduce amount of de-serializing work that needs to be done in python. By moving this straight into core we gain a lot of performance for python clients.

The API described above does not allow user to modify values in phi matrices. There will be another step (maybe in a separate pull request) that adds a new c_interface method to "attach" an external matrix to an existing topic model. Once this is done, the user will be able to manipulate values in the topic model directly. However, it won't be possible to change the structure of such matrix (e.g. to add or remove tokens or topics). It will be also not possible to "detach" the external matrix from topic model (the workaround to detach is to make a copy of the matrix via "MergeModel").


sashafrey:
Please, review the changes in python API. I've changed master.GetTopicModel and master.GetThetaMatrix to return a tuple, where first element is a protobuf message (TopicModel or ThetaMatrix), and the second element is a numpy matrix (phi or theta). The resulting protobuf message does not have the "bulk data" (e.g. TopicModel.topic_weights and ThetaMatrix.item_weights is empty).

This is a breaking change. Previous behavior is also available, but user needs to add "use_matrix=False" switch.

New functionality is also available for ProcessBatches, which can output theta matrix. However, by default ProcessBatches stores theta matrices into BigARTM cache, so that user can retrieve it later via GetThetaMatrix. An example of how to get numpy matrix from ProcessBatches is available in example15_import_export_topic_model.

Also note that enum ProcessBatchesArgs.ThetaMatrixType has changed (see messages.proto).

src/python/artm/library.py
```diff
@@ -564,7 +611,7 @@ def GetRegularizerState(self, regularizer_name):
         regularizer_state.ParseFromString(state_blob)
         return regularizer_state
 
-    def GetThetaMatrix(self, model=None, batch=None, clean_cache=None, args=None, topic_names=None):
+    def GetThetaMatrix(self, model=None, batch=None, clean_cache=None, args=None, topic_names=None, use_matrix=True):
```
MelLain: I have the only small question: why use_matrix default is True? It seems to broke the backward compatibility, doesn't it? False value is safe and if user need to speed up his code in low-level API, he will set it to True; the user of ArtmModel will has this option True by default in corresponding method implementation.

sashafrey: The idea was to promote this change as much as possible. I'm not 100% sure whether this is a good idea, but pros are that every BigARTM user (in python) will learn that now there is a new option of getting phi/theta as a numpy matrix. If they don't like it, it takes 30 seconds to fix their script by adding "use_matrix = False" (and that will be bold-highlighted in release notes). But I'm sure everyone will like it - it is more elegant and works substantially faster.

Typically I don't want to force users to upgrade, but this case sounds a bit special.

PR #246, opened 2015-07-17
Installation issues & python code structure

romovpa:
* Travis script cleanup
* Installation fixes in CMakeLists
* Python codebase moved outside `src/`
* Python codebase: added `setup.py` and appropriate CMakeLists.txt
* Python codebase: install python dependencies automatically
* Python API: now could import `artm.ArtmModel`
* Python API: artm.artm_model style fixes


CMakeLists.txt
```diff
@@ -92,5 +92,4 @@ if(("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") OR
 endif()
 
 add_subdirectory(src)
-
-install(FILES LICENSE DESTINATION .)
```
sashafrey: Current "install" steps are useful on Windows to create a release package. The copy all binaries, python api, examples, license and 3rdpartyy (protobuf/python) to "C:\Program Files\BigARTM". This makes it easy to prepare release drops for Windows. I only pack with 7z and upload to github :) Is it OK to apply your changes for all OS except windows, and keep the steps on windows as today?

python/artm/__init__.py
```diff
@@ -0,0 +1,18 @@
+from artm_model import (
```
sashafrey: Could you briefly explain why this change is useful? I'm really new to python so I simply don't know what to put in __init__.py... If there a good answer at stackoverflow or somewhere? I failed to find one...

src/python/artm_vs2012.sln
```diff
@@ -1,62 +0,0 @@
-﻿
```
sashafrey: I know @MelLain  is using Visual Studio as a tool to develop for python. Please, ask him if this is OK to remove the files. To me this is OK - I have switched to PyCharm, and I don't need any project files on disk.

MelLain: I don't see any reason to remove VS project. I prefer VS + pylint and dislike PyCharm. So, if it's not critical, I'd like the project file to present in a master branch.

python/artm/__init__.py
romovpa: - Example on StackOverflow: [why would I put python code in __init__.py files](http://stackoverflow.com/questions/5831148/why-would-i-put-python-code-in-init-py-files)
- Look at `__init__.py` of any `sklearn.<module>`, e.g. [sklearn.linear_model](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/linear_model/__init__.py)
- Look at any `__init__.py` in `gensim`, e.g. [gensim.models.__init__](https://github.com/piskvorky/gensim/blob/develop/gensim/models/__init__.py)

The general pattern consist in that you incapsulate all your auxiliary classes, functions, variables within the package in meaningfully separated py-modules. To make an interface for the user of your library, you add public classes and functions to the namespace of the package, this is done by imports in `__init__`.

src/python/artm_vs2012.sln
romovpa: I removed it because I have not VS installation and cannot correctly change those files according to new code structure. I suggest you to create those files and add them to the repository again, if you need them.

Nevertheless, I hate any kind of IDE files in public repositories. Not just me, see [StackOverflow explanation](http://stackoverflow.com/questions/14558450/why-ide-specific-project-files-are-ignored-in-git).

> I prefer VS + pylint and dislike PyCharm.

You prefer, but I cant use VS physically (I have not PC with Windows on board), what do you propose should I do? 

Many opensource projects add IDE files to `.gitignore`. See [sklearn](https://github.com/scikit-learn/scikit-learn/blob/master/.gitignore), [gensim](https://github.com/piskvorky/gensim/blob/develop/.gitignore).

Btw, we use CMake. Do you know the main destiny of it? CMake allows each developer setup its own environment and create config files for any IDE, see [wikipedia](https://en.wikipedia.org/wiki/CMake).

CMakeLists.txt
romovpa: Ok. Lets see, what it leads to in unix systems when you do the standard workflow ([`./configure && make && make install`](https://robots.thoughtbot.com/the-magic-behind-configure-make-make-install)).

Imagine I've build the project and I want to install BigARTM to the system:
```
root@0f1d2ab50075:/tmp/bigartm/build# make install
[  3%] Built target gflags-static
[  8%] Built target google-glog
[ 33%] Built target libprotobuf
[ 57%] Built target libprotoc
[ 58%] Built target protoc
[ 86%] Built target artm-static
[ 86%] Built target artm
[ 98%] Built target artm_tests
[100%] Built target cpp_client
Install the project...
-- Install configuration: "Release"
-- Installing: /usr/./LICENSE
-- Installing: /usr/bin/protoc
-- Set runtime path of "/usr/bin/protoc" to "$ORIGIN/../lib"
-- Installing: /usr/python/artm/library.py
-- Installing: /usr/python/artm/artm_model.py
-- Installing: /usr/python/artm/messages_pb2.py
-- Installing: /usr/python/artm/__init__.py
...
```
Stop, wtf BigARTM's `LICENSE` file do in my userspace dir, common for all the applications???

romovpa:
```
-- Installing: /usr/python/artm/library.py
-- Installing: /usr/python/artm/artm_model.py
-- Installing: /usr/python/artm/messages_pb2.py
-- Installing: /usr/python/artm/__init__.py
```
This is incorrect behaviour too. Because python in the user's environment could store packages not in `<install-prefix>/python`. After installation user needs to add this dir to PYTHONPATH, but this is a minor problem.

The main issue is that we've started to use python dependencies like `numpy` and `pandas`. Before installing we need to check them and install if it is needed. Dependency management in python is done with [PyPI](https://pypi.python.org/pypi) and `setup.py` scripts based on [distutils or setuptools](http://stackoverflow.com/questions/6344076/differences-between-distribute-distutils-setuptools-and-distutils2).

romovpa: I'm not about to remove `LICENSE` from installation, I mean we need to setup `CMakeLists` correctly.

Unfortunatelly, I am far from C++ project development and I have no way to master CMake. I'm just a stupid data scientist who promotes using [SciPy stack](http://www.scipy.org/about.html) in the work :(.

src/python/artm_vs2012.sln
MelLain: Ok, you convinced me:)

sashafrey:
Sorry for slow reply, I've been on vacation... :)

This pull request introduces many useful improvements, and I would like to bring them to master as soon as possible. However there are two issues:
1. This change weirdly fails on travis CI, and I can't understand why...
2. Changes in the install section of CMakeLists files break the way I create BigARTM release package for Windows. Yes, it is really important to fix all CMakeLists.txt so that "make install" works properly for Linux, but at the same time the change needs to be incremental, to maintain the way we install on Windows. If that's not possible then we need to re-design the way we deploy on Windows.

I suggest to split this change into independent parts, so that we can integrate to bigartm/master those changes that are ready. The following parts are fully independent and can be split into separate pull requests:

1. Fixes in artm_model.py  and __init__.py
2. Removal of visual studio files
3. Relocation src/python -> python/
   (needs to include minimal changes in travis.yml and CMakeLists.txt; installing on windows needs to work)
4. python/examples/download_datasets.sh (needs corresponding change in travis.yml)
5. other changes related to CMakeLists.txt and travis.yml.

Could you do the split into separate pull requests? I'll follow up on (3) and upgrade CMakeLists and verify they still work on windows. In parallel let's discuss how to move forward with (5). I suspect this discussion simpler once (1), (2), (3) and (4) are already in master, and the remaining fix is really small.

sashafrey:
Regarding travis - it does not run for your latest commit because of merge conflicts. You need to rebase to bigartm/master and re-push this branch, then CI will work again. I did this in https://github.com/bigartm/bigartm/pull/249, this is the way I see travis.ci is failing.

PR #251, opened 2015-07-26
Read from std::in when filename is "-" (hyphen)

sashafrey:


sashafrey:
appveyor build is not implemented yet, so failure is expected :) otherwise this change is ready for code review

PR #252, opened 2015-07-26
Appveyor

sashafrey:


src/artm/core/collection_parser.cc
```diff
@@ -24,6 +24,26 @@
 
 using boost::iostreams::mapped_file_source;
 
+namespace {
```
romovpa: Why those changes in this pull request?

sashafrey: My bad.. Fixed!

PR #256, opened 2015-08-04
Mel lain artm model fixes

MelLain:


codestyle_checks.sh
```diff
@@ -4,4 +4,7 @@ echo "Running C++ code checks"
 cat utils/cpplint_files.txt | xargs python utils/cpplint.py --linelength=120 || exit 1
 
 echo "Running Python code checks"
-pep8 --first --max-line-length=99 python/artm/artm_model.py || exit 1
+pep8 --first --max-line-length=99 python/artm/model.py || exit 1
```
romovpa:
```bash
for scr in python/artm/*.py
do
    pep8 --first --max-line-length=99 ${scr} || exit 1
done

File: python/artm/__init__.py [hunk 2]
```diff
@@ -1,13 +1,15 @@
-from artm_model import (
+from model import (
```
romovpa: Remove brackets

python/artm/__init__.py
```diff
@@ -1,13 +1,15 @@
...
     ArtmModel,
-
+)
+from regularizers import (
```
romovpa:
```python
from regularizers import *
```
and put `__all__` into the regularizers.py:
```python
__all__ = [
     'DecorrelatorPhiRegularizer',
     'ImproveCoherencePhiRegularizer',...
]
```

codestyle_checks.sh
MelLain: Recommended code will check library.py and messages_pb2.py files. First one will give errors, and second one is auto-generated. Conditions inside the loop will make the script too complex.

romovpa: Hm. Let's move `library.py` and `messages_pb2.py` into submodule `artm/library`?
Otherwise, 
`for scr in python/artm/{model,batches,regularizers,scores}.py`

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+                                                topic_names=topic_names)
+        tokens = [token for token in topic_model[0].token]
+        topic_names = [topic_name for topic_name in topic_model[0].topic_name]
+        retval = DataFrame(data=topic_model[1],
+                           columns=topic_names,
+                           index=tokens)
+
+        return retval
+
+    def get_theta(self, remove_theta=False):
```
romovpa: -> fit_transform

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+        theta_matrix = self.master.GetThetaMatrix(self.model, clean_cache=remove_theta)
+        document_ids = [item_id for item_id in theta_matrix[0].item_id]
+        topic_names = [topic_name for topic_name in theta_matrix[0].topic_name]
+        retval = DataFrame(data=theta_matrix[1].transpose(),
+                           columns=document_ids,
+                           index=topic_names)
+
+        return retval
+
+    def find_theta(self, batches=None, collection_name=None,
```
romovpa: -> transform

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+        """
+        if self.cache_theta is False:
+            raise ValueError("ArtmModel.get_theta(): cache_theta == False" +
+                             "Set ArtmModel.cache_theta = True")
+        if not self._initialized:
+            raise RuntimeError("ArtmModel.get_theta(): Model does not exist yet. Use " +
+                               "ArtmModel.initialize()/ArtmModel.fit_*()")
+
+        theta_matrix = self.master.GetThetaMatrix(self.model, clean_cache=remove_theta)
+        document_ids = [item_id for item_id in theta_matrix[0].item_id]
```
romovpa: dont use tuple as an array, instead:
```python
theta_message, theta_matrix = self.master.GetThetaMatrix(self.model, clean_cache=remove_theta)
```

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+        if not data_format == 'batches':
+            target_folder = tempfile.mkdtemp()
+        try:
+            batches_list = _parse_collection_inline(target_folder,
+                                                    data_path,
+                                                    data_format,
+                                                    'find_theta',
+                                                    collection_name,
+                                                    batches,
+                                                    batch_size=1000)
```
romovpa: Always name arguments when the list of them is long enough (number of arguments >= 3 then use names).
This prevents from an error when you add new arguments to the function.

python/artm/scores.py
```diff
@@ -0,0 +1,515 @@
...
+
+    _type = lib.ScoreConfig_Type_TopicKernel
+
+    def __init__(self, name=None, class_id=None, topic_names=None, eps=None,
+                 dictionary_name=None, probability_mass_threshold=None):
+        BaseScore.__init__(self,
+                           name=name,
+                           class_id=class_id,
+                           topic_names=topic_names)
+        self._eps = GLOB_EPS
```
romovpa: group lines by assigning field:
```python
self._eps = GLOB_EPS
if eps is not None:
    self._config.theta_sparsity_eps = eps
    self._eps = eps

self._dictionary_name = ''
if dictionary_name is not None:
      ...
```

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+        self._initialized = True
+        topic_model = self.master.GetTopicModel(model=self.model, use_matrix=False)
+        self._topic_names = [topic_name for topic_name in topic_model.topic_name]
+        self._num_topics = topic_model.topics_count
+
+        # Remove all info about previous iterations
+        self._scores_info = {}
+        self._synchronizations_processed = 0
+
+    def to_csv(self, file_name='artm_model.csv'):
```
romovpa: remove
(implemented in pandas.DataFrame)

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+        from the lib
+
+        Args:
+          dictionary_name (str): the name of the dictionary in th lib, default=None
+        """
+        if dictionary_name is not None:
+            self.master.lib_.ArtmDisposeDictionary(self.master.id_, dictionary_name)
+        else:
+            raise IOError('ArtmModel.remove_dictionary(): dictionary_name is None')
+
```
romovpa: add general `fit()` method

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+            batch_size (int): number of documents to be stored ineach batch,
+            default=1000
+
+        Note:
+          ArtmModel.initialize() should be proceed before first call
+          ArtmModel.fit_offline(), or it will be initialized by dictionary
+          during first call.
+        """
+        if collection_name is None and data_format == 'bow_uci':
+            raise IOError('ArtmModel.fit_offline(): No collection name was given')
```
romovpa: Dont add the name of method to the exception message, it is obvious from traceback.

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+
+        if not data_format == 'batches' and batches is not None:
+            raise IOError("ArtmModel.fit_offline(): batches != None" +
```
romovpa: This is not IOError. Why are you care about this situation?
I think this check is not needed, if you receive batches when use other data format it is not a problem, just ignore batches.

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+
+    # ========== CONSTRUCTOR ==========
+    def __init__(self, num_processors=0, topic_names=None, num_topics=10,
+                 class_ids=None, num_document_passes=10, cache_theta=True):
+        self._num_processors = 0
+        self._num_topics = 10
+        self._num_document_passes = 10
+        self._cache_theta = True
+
+        if topic_names is None or topic_names is []:
```
romovpa: http://stackoverflow.com/questions/53513/best-way-to-check-if-a-list-is-empty

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+        Returns:
+          pandas.DataFrame: (data, columns, rows), where:
+          1) columns --- the ids of documents, for which the Theta
+          matrix was requested
+          2) rows --- the names of topics in topic model, that was
+          used to create Theta
+          3) data --- content of Theta matrix
+        """
+        if self.cache_theta is False:
+            raise ValueError("ArtmModel.get_theta(): cache_theta == False" +
```
romovpa: Dont use method name in the message

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+
+            results = self.master.ProcessBatches(
+                pwt=self.model,
+                batches=batches_list,
+                target_nwt='nwt_hat',
+                inner_iterations_count=self._num_document_passes,
+                class_ids=self._class_ids,
+                theta_matrix_type=lib.ProcessBatchesArgs_ThetaMatrixType_External)
+
+            document_ids = [item_id for item_id in results[0].theta_matrix.item_id]
```
romovpa: Dont get tuple elements by index, use tuple as tuple!

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+            topic_names = [topic_name for topic_name in results[0].theta_matrix.topic_name]
+            retval = DataFrame(data=results[1].transpose(),
```
romovpa: retval?
Shtoza nafig retval?

`df`
`theta_frame`
`theta_df`

python/artm/scores_info.py
```diff
@@ -0,0 +1,724 @@
+import collections
```
romovpa: `score_info.py`

python/artm/scores_info.py
```diff
@@ -0,0 +1,724 @@
...
+          list of int: total number of rows in Phi on synchronizations
+        """
+        return self._total_tokens
+
+    @property
+    def last_value(self):
+        """Returns:
+        double: value of Phi sparsity on the last synchronization
+        """
+        return self._value[len(self._value) - 1]
```
romovpa:
```python
return self._value[-1]
```

python/artm/scores_info.py
```diff
@@ -0,0 +1,724 @@
...
+          in direct order of document_ids
+        """
+        return self._snippet
+
+    @property
+    def last_document_ids(self):
+        """Returns:
+          list of int: ids of documents in snippet on the last synchronization
+        """
+        return self._document_ids
```
romovpa: 720 lines on python,
My brain hates you!

```python
class ScoreTracker(object):

    _TRACK_FIELDS = []

    def __init__(self, score):
        self.name = score.name
        self.tracks = {field: [] for field in self._TRACK_FIELDS}

    def __getattr__(self, attr_name):
        if attr_name in self.tracks:
            return self.tracks[attr_name]

        if attr_name.startswith('last_'):
            attr_name = attr_name[len('last_'):]
            if attr_name in self.tracks:
                return self.tracks[-1]

        raise AttributeError(attr_name)


    def add(self, score):
        for field in self._TRACK_FIELDS:
            value = getattr(score, field) if score is not None else None
            self.tracks[field].append(value)


class SparsityThetaTracker(ScoreTracker):
    _TRACK_FIELDS = [
        'value',
        'zero_topics',
        'total_topics',
    ]

class PerplexityTracker(ScoreTracker):
    _TRACK_FIELDS = [
        'value',
        'raw',
        'normalizer',
        'zero_tokens',
        'theta_sparsity_value',
        'theta_sparsity_zero_topics',
        'theta_sparsity_total_topics',
    ]

class ItemsProcessedScoreInfo(ScoreTracker):
    _TRACK_FIELDS = [
        'value',
    ]

...
```

python/artm/scores_info.py
romovpa: I dont like the notion "Score Info".
May be `ScoreTracker`?

python/artm/scores.py
```diff
@@ -0,0 +1,515 @@
...
+
+CREATE_SCORE_CONFIG = {
+    lib.ScoreConfig_Type_SparsityPhi: messages_pb2.SparsityPhiScoreConfig,
+    lib.ScoreConfig_Type_SparsityTheta: messages_pb2.SparsityThetaScoreConfig,
+    lib.ScoreConfig_Type_Perplexity: messages_pb2.PerplexityScoreConfig,
+    lib.ScoreConfig_Type_ItemsProcessed: messages_pb2.ItemsProcessedScoreConfig,
+    lib.ScoreConfig_Type_TopTokens: messages_pb2.TopTokensScoreConfig,
+    lib.ScoreConfig_Type_ThetaSnippet: messages_pb2.ThetaSnippetScoreConfig,
+    lib.ScoreConfig_Type_TopicKernel: messages_pb2.TopicKernelScoreConfig
+}
```
romovpa: More concise is to add field `_config_message` to each class:
```python
class SparsityPhiScore(BaseScore):
    """SparsityPhiScore is a score in ArtmModel (public class)
    """

    _type = lib.ScoreConfig_Type_SparsityPhi
    _config_message = messages_pb2.SparsityPhiScoreConfig
```

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+        self._model = 'pwt'
+        self._regularizers = regularizers.Regularizers(self.master)
+        self._scores = scores.Scores(self.master, self._model)
+
+        self._scores_info = {}
+        self._synchronizations_processed = 0
+        self._initialized = False
+
+    # ========== PROPERTIES ==========
+    @property
```
romovpa: Hate those properties! Properties! Properties everywhere!!!

python/artm/model.py
sashafrey: ``retval`` stands for *return value*, and it is is quite common naming convention in C++:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa367158(v=vs.85).aspx
In addition there are things like ``rhs`` (right-hand-size) or ``other``, which are commonly used as an argument of copy constructors / ``operator=``.

If these are similar abbreviation are not common in Python we should indeed stay away from them...

python/artm/regularizers.py
```diff
@@ -0,0 +1,421 @@
...
+            return self._data[name]
+        else:
+            raise KeyError('No regularizer with name ' + name)
+
+    @property
+    def data(self):
+        return self._data
+
+
+class BaseRegularizer(object):
```
romovpa: - each field is duplicated: object field and field of the protobuf message
- lot of meaningless properties, copied many times 

... :(

I want a bit sophisticated base class:
```python
class BaseRegularizer(object):

    _TYPE = None
    _CONFIG_MESSAGE = None

    def __init__(self, name, tau, topic_names, **kwargs):
        if _TYPE is None or _CONFIG_MESSAGE is None:
            raise NotImplementedError('{} is abstract'.format(self.__class__))

        self.name = name if name is not None else '{}: {}'.format(self._TYPE, uuid.uuid1().urn)
        self.tau = tau

        self.config = self._CONFIG_MESSAGE()
        for field_name, value in kwargs.iteritems():
            setattr(self.config, field_name, value)
            # TODO: handle complex fields, use the self._FIELDS description

    def __getattr__(self, attr_name):
        if hasattr(self.config, attr_name):
            return getattr(self.config, attr_name)
        else:
            raise AttributeError(attr_name)

    def __setattr__(self, attr_name, value):
        if hasattr(self.config, attr_name):
            setattr(self.config, attr_name, value)
            self.lib_.Reconfigure(config)
        else:
            raise AttributeError(attr_name)
```

And a very concise and clean regularizer classes, like this:
```
class SmoothSparsePhiRegularizer(BaseRegularizer):
    """SmoothSparsePhiRegularizer is a regularizer in ArtmModel (public class)
    """

    _TYPE = lib.RegularizerConfig_Type_SmoothSparsePhi
    _CONFIG_MESSAGE = messages_pb2.SmoothSparsePhiConfig
    _FIELDS = [
        'topics',
        'namespaces',
        'dictionary_name',
    ]
```

or maybe with django-styled fields:
```python
class SmoothSparsePhiRegularizer(BaseRegularizer):
    """SmoothSparsePhiRegularizer is a regularizer in ArtmModel (public class)
    """

    _TYPE = lib.RegularizerConfig_Type_SmoothSparsePhi
    _CONFIG_MESSAGE = messages_pb2.SmoothSparsePhiConfig
    
    topics = fields.StringListFields()
    namespaces = fields.StringListFields()
    dictionary_name = fields.PathField()

```

python/artm/regularizers.py
```diff
@@ -0,0 +1,421 @@
...
+]
+
+CREATE_REGULARIZER_CONFIG = {
+    lib.RegularizerConfig_Type_SmoothSparsePhi: messages_pb2.SmoothSparsePhiConfig,
+    lib.RegularizerConfig_Type_SmoothSparseTheta: messages_pb2.SmoothSparseThetaConfig,
+    lib.RegularizerConfig_Type_DecorrelatorPhi: messages_pb2.DecorrelatorPhiConfig,
+    lib.RegularizerConfig_Type_LabelRegularizationPhi: messages_pb2.LabelRegularizationPhiConfig,
+    lib.RegularizerConfig_Type_SpecifiedSparsePhi: messages_pb2.SpecifiedSparsePhiConfig,
+    lib.RegularizerConfig_Type_ImproveCoherencePhi: messages_pb2.ImproveCoherencePhiConfig
+}
```
romovpa: Place this information inside the regularizer classes.

python/artm/batches_utils.py
```diff
@@ -0,0 +1,68 @@
+import os
+
+import messages_pb2
+import library
```
romovpa: It is deprecated. Isn't it?

romovpa: Always use relative imports for touching local modules:
```python
from . import library
from artm import library
```

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+
+    elif data_format == 'plain_text':
+        raise NotImplementedError()
+    else:
+        raise IOError('ArtmModel.' + method_name + '(): Unknown data format')
+
+    return batches_list
+
+
+class ArtmModel(object):
```
romovpa: Rename to `ARTM`, seriously

python/artm/model.py
```diff
@@ -0,0 +1,723 @@
...
+        if not self._initialized:
+            raise RuntimeError("Model does not exist yet. Use " +
+                               "ArtmModel.initialize()/ArtmModel.fit_*()")
+
+        topic_model = self.master.GetTopicModel(model=self.model,
+                                                class_ids=class_ids,
+                                                topic_names=topic_names)
+        tokens = [token for token in topic_model[0].token]
+        topic_names = [topic_name for topic_name in topic_model[0].topic_name]
+        retval = DataFrame(data=topic_model[1],
```
romovpa: retval... indexing the tuple...

python/artm/model.py
romovpa: Awful convention!
`retval` says nothing about variable content.
It is obvious with `return` construction that the variable is a value that returns. But one cant even understand what the function value is. The code says: "this function returns the returning value", awesom.

python/artm/scores.py
```diff
@@ -0,0 +1,515 @@
...
+            self._config.eps = eps
+            self._eps = eps
+
+    @property
+    def eps(self):
+        return self._eps
+
+    @property
+    def class_id(self):
+        raise KeyError('No class_id parameter')
```
romovpa: Very strange. Are you defined this property in the base class to disable it in the successor?

romovpa:
I propose to write a micro-framework for wrapping reconfigurable BigARTM objects and use it to remove meaningless code from the implementation of regularizers, scorers and even the model class. This could be implemented using [python metaclasses](http://eli.thegreenplace.net/2011/08/14/python-metaclasses-by-example):

```python
## Technical stuff, utils.py

class _ConfigurableClass(type):
    """Metaclass that creates classes of all reconfigurable BigARTM objects (models, regularizers, scorers)"""
    def __new__(name, bases, dct):
        raise NotImplementedError()
    def __init__(cls, name, bases, dct):
        raise NotImplementedError()

class Field(object):
    """Describes all the information needed to create the field of configurable class"""
    def __init__(self, type, setter=False, default=None):
        raise NotImplementedError()

## Scorers

class BaseScore(object):
    __metaclass__ = _ConfigurableClass

    ConfigMessage = None
    TypeId = None

    name = Field(type=str)
    topics = Field(type=list, setter=True)


class SparsityPhiScore(BaseScore):
    ConfigMessage = messages_pb2.SparsityPhiScoreConfig
    TypeId = lib.ScoreConfig_Type_SparsityPhi

    eps = Field(type=float, setter=True)


class SparsityThetaScore(BaseScore):
    ConfigMessage = messages_pb2.SparsityThetaScoreConfig
    TypeId = lib.ScoreConfig_Type_SparsityTheta

    eps = Field(type=float, setter=True)
    namespaces = Field(type=list, setter=True)

...

# Regularizers

class BaseRegularizer(object):
    __metaclass__ = _ConfigurableClass

    name = Field(type=str)
    topics = Field(type=list, setter=True)

class BaseRegularizerPhi(BaseRegularizer):
    namespaces = Field(type=list, setter=True)
    dictionary_name = Field(type=str)


class DecorrelatorPhiRegularizer(BaseRegularizerPhi):
    ConfigMessage = messages_pb2.DecorrelatorPhiConfig
    TypeId = lib.RegularizerConfig_Type_DecorrelatorPhi

...

# Model

class ARTM(sklearn.base.BaseEstimator, sklearn.base.TransformerMixin):
    __metaclass__ = _ConfigurableClass

    num_processors = Field(type=int, setter=True)
    topics = Field(type=list, setter=True)
    num_document_passes = Field(type=int, default=10, setter=True)
    cache_theta = Field(type=bool, default=False, setter=True)
    ...

    def fit(self, collection, **kwargs):
        self.fit_online(...)
        return self

    def fit_offline(self, collection, passes=10):
        raise NotImplementedError()

    def fit_online(self, collection, alpha=0.1, update_every=100):
        raise NotImplementedError()

    def transform(self, collection):
        theta = self._get_theta(collection)
        return theta

    def fit_transform(self, collection, **fit_params):
        previous_cache_theta = self.cache_theta
        self.cache_theta = True
        self.fit(collection, **fit_params)
        theta = self._get_cached_theta()
        self.cache_theta = previous_cache_theta
        return theta

    def save(self, filename):
        raise NotImplementedError()

    def load(self, filename):
        raise NotImplementedError()

    @property
    def phi_(self):
        return self._get_phi()
```

python/artm/model.py
sashafrey: Not really... It should be sufficient to read the signature of the function to understand what it returns. Its really bad if you need to go into implementation to understand what you will get as an output. So I like ``retval`` for C++. For Python this is getting worse because the type of the returned value is not part of function signature.

python/artm/model.py
sashafrey: I fully agree on this, but I would like to comment on some differences between C++ vs Python here. 

For native C++ I prefer to add the method name into the exception because call stack is typically not available (and its nearly impossible to get the call stack in a cross-platform fashion). The only case when native call is easily available are crashes :)
Also, note that ``c_interface`` has a translation layer that maps C++ exceptions into an integer error code, and sets last error message. Then this error code is converted into a python exception by the wrapper. No native call stack info is passed from C++ to Python at this point, so all info should be in the exception message itself.

python/artm/scores_info.py
sashafrey: Wow! Cool stuff :+

python/artm/model.py
sashafrey: Agree, let it be ``ARTM``

sashafrey:
Agree, this looks promising, and far easier to maintain. 
One question on 
```
class ARTM(sklearn.base.BaseEstimator, sklearn.base.TransformerMixin):
```
Do you really want to inherit this from ``sklearn.base.BaseEstimator`` and ``TransformerMixin``? What if user doesn't have sklearn? Do we require this as a dependency?

romovpa:
> Do you really want to inherit this from sklearn.base.BaseEstimator and TransformerMixin? What if user doesn't have sklearn? Do we require this as a dependency?

Very good point. In this case... duck typing is enough.

python/artm/batches_utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+DICTIONARY_NAME = 'dictionary'
+
+
+class Batch(object):
+    def __init__(self, file_name):
+        self._file_name = file_name
+
+    def __str__(self):
+        return self._file_name
```
romovpa:
```python
def __repr__(self):
    return 'Batch({})'.format(self._file_name)
```

python/artm/batches_utils.py
```diff
@@ -0,0 +1,96 @@
...
+from . import wrapper
+from wrapper import constants
+from wrapper import messages_pb2 as messages
+
+DICTIONARY_NAME = 'dictionary'
+
+
+class Batch(object):
+    def __init__(self, file_name):
+        self._file_name = file_name
```
romovpa: I recommend to make the path absolute asap:
```python
self._filename = os.path.abspath(filename)
```

romovpa: `file_name` —> `filename`

A list of my arguments: [[1]](http://english.stackexchange.com/questions/5366/which-is-correct-filename-file-name-or-filename), [[2]](https://en.wikipedia.org/wiki/Filename), [[3]](https://docs.python.org/2/library/os.path.html) (they're using "filename" in the text)

python/artm/batches_utils.py
```diff
@@ -0,0 +1,96 @@
...
+                    raise RuntimeError('No batches were found')
+            else:
+                self._batches_list = [Batch(os.path.join(data_path, batch)) for batch in batches]
+
+        elif data_format == 'bow_uci' or data_format == 'vowpal_wabbit':
+            parser_config = messages.CollectionParserConfig()
+            parser_config.num_items_per_batch = batch_size
+            if data_format == 'bow_uci':
+                parser_config.docword_file_path =\
+                    os.path.join(data_path, 'docword.' + collection_name + '.txt')
```
romovpa: [PEP8](https://www.python.org/dev/peps/pep-0008/#id14):
> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

```python
parser_config.docword_file_path = os.path.join(
    data_path, 'docword.' + collection_name + '.txt')
```

romovpa: Stop using string concatenation!
```python
os.path.join(data_path, 'docword.{}.txt'.format(collection_name))
```
or
```python
os.path.join(data_path, 'docword.%s.txt' % collection_name)
```

Concatenation is much harder to understand.

python/artm/master_component.py
```diff
@@ -287,91 +247,166 @@ def attach_model(self, model, master_id=None):
...
+        elif score_data.type == constants.ScoreData_Type_SparsityPhi:
+            score_info = messages.SparsityPhiScore()
+        elif score_data.type == constants.ScoreData_Type_ItemsProcessed:
+            score_info = messages.ItemsProcessedScore()
+        elif score_data.type == constants.ScoreData_Type_TopTokens:
+            score_info = messages.TopTokensScore()
+        elif score_data.type == constants.ScoreData_Type_ThetaSnippet:
+            score_info = messages.ThetaSnippetScore()
+        elif score_data.type == constants.ScoreData_Type_TopicKernel:
+            score_info = messages.TopicKernelScore()
```
romovpa: I feel the hellraising of the `artm.library`, woohoha :imp:

python/artm/model.py
```diff
@@ -0,0 +1,583 @@
...
+          dictionary_name (str): the name of the dictionary in th lib, default=None
+        """
+        if dictionary_name is not None:
+            self._lib.ArtmDisposeDictionary(self.master.master_id, dictionary_name)
+        else:
+            raise IOError('dictionary_name is None')
+
+    def fit_offline(self, batch_vectorizer=None, num_collection_passes=1, decay_weight=0.0,
+                    apply_weight=1.0, reset_theta_scores=False):
+        """ARTM.fit_offline() --- proceed the learning of
```
romovpa: It is not necessary to write the name of class and the method in the docstring, bacause you will easily see the whole signature in the documentation.
```python
In [3]: from sklearn import decomposition
In [5]: ?decomposition.NMF.fit
Signature: decomposition.NMF.fit(self, X, y=None, **params)
Docstring:
Learn a NMF model for the data X.
...
```

python/artm/regularizers.py
```diff
@@ -0,0 +1,423 @@
...
+            raise KeyError('No regularizer with name ' + name)
+
+    @property
+    def data(self):
+        return self._data
+
+
+class BaseRegularizer(object):
+    def __init__(self, name, tau, topic_names):
+        config = self._config_message()
```
romovpa: Put empty `_config_message` into the base class and add the check:
```python
if self._config_message is None:
    raise NotImplementedError()
config = self._config_message()
```

python/artm/batches_utils.py
```diff
@@ -0,0 +1,96 @@
...
+      batches(list of str): list with non-full file names of batches (necessary parameters are
+      batches + data_path + data_fromat=='batches' in this case)
+    """
+    def __init__(self, batches=None, collection_name=None, data_path='', data_format='batches',
+                 target_folder='', batch_size=1000, dictionary_name=DICTIONARY_NAME):
+        self._batches_list = []
+        if data_format == 'batches':
+            if batches is None:
+                batch_str = glob.glob(os.path.join(data_path, '*.batch'))
+                self._batches_list = [Batch(batch) for batch in batch_str]
```
romovpa:
```python
batch_filenames = glob.glob(os.path.join(data_path, '*.batch'))
self._batches_list = [Batch(filename) for filename in batch_filenames]
```

python/artm/regularizers.py
```diff
@@ -0,0 +1,423 @@
...
+
+        self._probability_threshold = 0.99
+        if probability_threshold is not None:
+            self._config.probability_threshold = probability_threshold
+            self._probability_threshold = probability_threshold
+
+        self._sparse_by_columns = True
+        if sparse_by_columns is not None:
+            if sparse_by_columns is True:
+                self._config.mode = lib.SpecifiedSparsePhiConfig_Mode_SparseTopics
```
romovpa: Hmm... `lib`?

python/artm/regularizers.py
```diff
@@ -0,0 +1,423 @@
...
+
+    @sparse_by_columns.setter
+    def sparse_by_columns(self, sparse_by_columns):
+        self._sparse_by_columns = sparse_by_columns
+        config = messages.SpecifiedSparseRegularizationPhiConfig()
+        config.CopyFrom(self._config)
+        if sparse_by_columns is True:
+            config.mode = lib.SpecifiedSparsePhiConfig_Mode_SparseTopics
+        else:
+            config.mode = lib.SpecifiedSparsePhiConfig_Mode_SparseTokens
```
romovpa: `lib` is not defined here

romovpa:
I'm okay. Ship it! :shipit:

sashafrey:
https://github.com/MelLain/bigartm/commit/aa3064e8a1a154f1a420dffa1c38a48ccd82aede
Could you replace this commit with rebase?

MelLain:
Was moved to Mel lain python artm #269

PR #258, opened 2015-08-04
Read from std::in when filename is "-" (hyphen)

sashafrey:


src/artm/core/collection_parser.cc
```diff
@@ -24,6 +24,26 @@
 
 using boost::iostreams::mapped_file_source;
 
+namespace {
+class ifstream_or_cin {
```
romovpa: Hey, what the difference between `ifstream` and `cin`?
It seemed to me that `cin` is a special instance of `ifstream` (or maybe just `istream`). In unix systems there are a special filenames for cin/cout/cerr: `/dev/stdin`, ...

romovpa:
travis have failed

sashafrey:
ci is hopefully fixed - I forget to include <iostream>

std::cin and std::cout are part of the C++ standard
http://www.cplusplus.com/reference/iostream/cin/
They are cross-platform, while /dev/stdin won't work on windows.

sashafrey:
``ifstream`` is an input stream class to operate on files
http://www.cplusplus.com/reference/fstream/ifstream/
The problem is that standard input is not a file, so I'm not sure ``ifstream`` can open it (unless OS provide a special hook for it like ``/dev/stdin``

romovpa:
:ok_hand:

PR #259, opened 2015-08-04
Restore CMakeLists for BigARTM installation on Windows

sashafrey:


sashafrey:
@romovpa I know that's ugly to keep so different Install sections for CMakeLists under Linux and Windows, but may we live with this for now? There are several goodies that CMakeLists does for us in terms of "installing" BigARTM under Windows. Particularly, it finds correct Visual Redistributable version, and copies the binaries into the target folder. This works for any version of Visual Studio. It will take some time to automate the same behavir  via .bat or .ps1.
Same thing applies to x64 vs Win32 version -- they automatically go into correct folders (Program Files or Program Files (x86)).
Plus its nice to keep installation steps in CMakeLists because then I can use things like
```
install(TARGETS artm DESTINATION bin)
```
Here I don't care how CMake finds correct artm.dll (not even that the extension is .dll). That's fairly convenient.

romovpa:
`make install` on ubuntu does this:
```
Install the project...
-- Install configuration: "Release"
-- Installing: /usr/bin/protoc
-- Set runtime path of "/usr/bin/protoc" to "$ORIGIN/../lib"
-- Installing: /usr/lib/libartm.so
-- Installing: /usr/bin/cpp_client
```

ok.

Don't like installation of imposed `protoc` binary of BigARTM-specific version. 
It's very bad in general, e.g. when you try to install BigARTM to your machine it overrides your protoc with bigartm-prefered version (hate this behaviour!). But when you use docker container, it's ok.

romovpa:
As a temporary solution — ok.

sashafrey:
Protoc is installed in
https://github.com/bigartm/bigartm/blob/master/3rdparty/protobuf-cmake/protoc/CMakeLists.txt
Should I put this inside ``if(msvc)`` ?

romovpa:
> Should I put this inside if(msvc) to prevent this for linux?

yes, please

sashafrey:
done

romovpa:
:ok_hand:

PR #260, opened 2015-08-05
Python low-level wrapper

romovpa:
Simple binding of Plain C Interface in Python. Automatically wraps all the C-style functions from the core BigARTM API to python-friendly functions.

Proposed package artm.wrapper is considered as a replacement for the old hard-maintainable artm.library module.


sashafrey:
I agree with the goal - artm.library needs to be replaced. We should keep it for a while to give people a chance to switch their scripts to either ArtmModel or to this low-level API. Over time artm.library can be removed completely.

In this pull request:
* wrapper lacks ``ArtmAttachModel`` and ``ArtmRequestResultEx``, plus some other methods. Please, check [c_interface](https://github.com/bigartm/bigartm/blob/master/src/artm/c_interface.h) for the entire list. The [documentation](http://docs.bigartm.org/en/latest/ref/c_interface.html) is incomplete...
* it would be nice to add simple tests to verify that the API works

romovpa:
All new api call added to the wrapper.

I feel dissonance in naming of `ArtmRequestResult` and `ArtmRequestResultEx`. I propose to consider renaming:
 - `ArtmRequestResultEx` -> `ArtmCopyRequestedObject` (bacause it can copy arbitrary object, not only ndarrays)
 - `ArtmRequestResult` -> `ArtmCopyRequestedMessage`

romovpa:
Another thing that annoys me:
The call `ArtmRequestResult` is tied with another request-call with the returned message length, whereas the call `ArtmRequestResultEx` is not actually tied with the previous call. In the wrapper the `ArtmRequestResult` call is implemented with internal `_get_requested_message()` method because there are no way to implement it as an ordinary call (because of returned length of the requested message). The `ArtmRequestResultEx` is implemented as an ordinary call and could be executed or not.

sashafrey:
We agreed that the all API renames (such as  `ArtmRequestResult` -> `ArtmCopyRequestedMessage`) will be listed in a dedicated wiki page. We will then schedule then and do all in one go, after notifying users that this is going to happen.

PR #261, opened 2015-08-05
Batch weights

sashafrey:
This change upgrades topic_count (an integer field) into topic_weight (a newly introduced  float field). This is an important scenario for some matrix decompositions where original matrix does represent something else rather then integer n_dw counters.

This change also introduces a "repeated float batch_weight" field in ProcessBatchesArgs. This allows users to specify different weight for each batch. This works as if all token_weight values in the batch had been multiplied by batch_weights. This has effect only in model inference. All scores, such as perplexity, currently ignore batch_weights field.


sashafrey:
This change upgrades topic_count (an integer field) into topic_weight (a newly introduced  float field). This is an important scenario for some matrix decompositions where original matrix does represent something else rather then integer n_dw counters.

This change also introduces a "repeated float batch_weight" field in ProcessBatchesArgs. This allows users to specify different weight for each batch. This works as if all token_weight values in the batch had been multiplied by batch_weights. This has effect only in model inference. All scores, such as perplexity, currently ignore batch_weights field.

sashafrey:
@MelLain Does this change look OK ?

romovpa:
Could you list what is changed for the user? What interfaces have been changed?

MelLain:
Change looks correct, OK.

sashafrey:
@romovpa Chances from user perspective are described in the comment of this pull request...

sashafrey:
API chances:
- new protobuf fields Field.token_weight, ProcessBatchesArgs.batch_weight, DictionaryEntry.token_weight, DictionaryConfig.total_token_weight
- obsoleted fields: Field.token_count, DictionaryConfig.total_token_count,  DictionaryEntry.token_count. All obsoleted fields are automatically upgraded to new fields, so this change is backwards compatible.
- no chances in C API.
- Minimal chances in library.py to let user specify batch_weight without creating protobuf message

sashafrey:
Ah, right.. One more API change :)
Now we support float numbers in the 3rd column of the docword file. I'll check if this is consistent with our VW parser -- I guess we should now allow non-integer values there as well.

sashafrey:
Yes, we do. This change also upgrades VW parser to handle things like
```
doc2 |author noname |@default_class hello:5.1
```
(note 5.1 at the end - I mean this is float). This change does not address any localization issues, because this is a bit challenging in C++. I think BigARTM always expects a period as decimal separator (not coma). There is a room for improvement here, but this would be out of scope for this pull request.

romovpa:
Aha! 
@sashafrey Thanks, I will consider this when rethinking ArtmModel

PR #262, opened 2015-08-06
Retrieve external matrix layout via dedicated APIs

sashafrey:
This pull request adds three API methods in c_interface:

``ArtmRequestProcessBatchesExternal`` (an alternative to ``ArtmRequestProcessBatches``)
``ArtmRequestThetaMatrixExternal`` (an alternative to ``ArtmRequestThetaMatrix``)
``ArtmRequestTopicModelExternal`` (an alternative to ``ArtmRequestTopicModel``)

The ``ArtmRequestXxx`` request 1 follow up call to ``ArtmRequestResult``.
The ``ArtmRequestXxxExternal``  require 2 follow up calls: first to ``ArtmRequestResult``, and then to ``ArtmRequestResultEx``.

Until this fix both behaviours were hidden in one function (``ArtmRequestXxx``).
The argument (``GetTopicModelArgs``, ``GetThetaMatrixArgs``, or ``ProcessBatchesArgs``) was used to distinguish two behaviours.

This fix makes a well-defined contract for ``ArtmRequestXxx`` and ``ArtmRequestXxxExternal`` and removes ``External`` matrix layouts from all protobuf messages.


sashafrey:
@romovpa This is what we discussed yesterday. I have only fixed stuff in c_interface, not the low-level python api (library.py). E.g. it some methods in library.py still return protobuf message OR tuple(protobuf,numpy) depending on use_matrix flag. This is wrong, but I don't think it worse changing this given that you are working on a new python wrapper for c_api.

romovpa:
:+1: 
Merge?

PR #263, opened 2015-08-10
init test for wrapper

MelLain:


python/wrapper_tests/download_datasets.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
romovpa: from travis:
```
/home/travis/build.sh: line 41: ./download_datasets.sh: Permission denied
```
run 
```bash
chmod +x download_datasets.sh
```
and re-commit to make it executable

.travis.yml
```diff
@@ -36,10 +37,12 @@ install:
...
 before_script:
   - pushd python/examples && ./download_datasets.sh && popd
+  - pushd python/wrapper_tests && chmod +x download_datasets.sh && popd
 
 script:
   - export PYTHONPATH=`pwd`/python:$PYTHONPATH
   - export ARTM_SHARED_LIBRARY=`pwd`/build/src/artm/libartm.so
   - pushd build/src/artm_tests && ./artm_tests && popd
-  - pushd python/tests && python tests.py && popd
+  - pushd python/library_tests && python tests.py && popd
```
romovpa: I prefer one folder for all tests:
```
python/
    tests/artm/
        wrapper/ 
            test01_xxx.py
            test02_yyy.py
        model/ (for ArtmModel)

File: .travis.yml [hunk 3]
```diff
@@ -36,10 +37,12 @@ install:
 
 before_script:
   - pushd python/examples && ./download_datasets.sh && popd
+  - pushd python/wrapper_tests && chmod +x download_datasets.sh && popd
```
romovpa: no chmod in ci-script! think about users! commit file with the x-bit to repo

python/wrapper_tests/download_datasets.sh
```diff
@@ -0,0 +1,6 @@
...
+
+wget https://s3-eu-west-1.amazonaws.com/artm/docword.kos.txt.gz
+gunzip docword.kos.txt.gz
```
romovpa: gzip -d

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+import uuid
+import shutil
+import os
+import pytest
+
+def test_func():
+    # Set some constants
+    num_tokens = 60
+    num_items = 100
+    batches_folder = os.getcwd() + '/temp_batches'
```
romovpa: os.path.abspath('temp_batches')

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+    pwt = 'pwt'
+    nwt = 'nwt'
```
romovpa: wtf??

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # Generate small collection of random items
+    batch = messages.Batch()
+    batch.id = str(uuid.uuid4())
+    for token_id in xrange(num_tokens):
+        batch.token.append('token' + str(token_id))
```
romovpa: Never use string concatenation, string interpolation instead:
```python
batch.token.append('token_%s' % token_id)
```
or more concise:
```python
batch.token.append('token_{id}'.format(id=token_id))
```

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+
+    for item_id in xrange(num_items):
+        item = batch.item.add()
+        item.id = item_id
+        field = item.field.add()
+        for token_id in range(0, num_tokens):
+            field.token_id.append(token_id)
+            background_count = random.randint(1, 5) if (token_id >= 40) else 0
```
romovpa: hmm... random in tests??

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+            topical_count = 10 if (token_id < 40) and ((token_id % 10) == (item_id % 10)) else 0
+            field.token_count.append(background_count + topical_count)
+
+    # Create the instance of low-level API
+    lib = artm.wrapper.LibArtm()
+
+    # Save batch on the disk
+    lib.ArtmSaveBatch('temp_batches', batch)
```
romovpa: use `tempfile` module to get temporary file name

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+import os
+import pytest
+
+def test_func():
+    # Set some constants
+    num_tokens = 60
+    num_items = 100
+    batches_folder = os.path.join(os.getcwd() + '/temp_batches')
+    pwt = 'pwt'
+    nwt = 'nwt'
```
romovpa: `x = 'x'`
wtf??

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+import uuid
+import shutil
+import os
+import pytest
+
+def test_func():
+    # Set some constants
+    num_tokens = 60
+    num_items = 100
+    batches_folder = os.path.join(os.getcwd() + '/temp_batches')
```
romovpa: usin `tempfile.mkdtemp()` is a correct way to create temporary directory, and remember to remove it at the end

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+    batch = messages.Batch()
+    batch.id = str(uuid.uuid4())
+    for token_id in xrange(num_tokens):
+        batch.token.append('token' + str(token_id))
+
+    for item_id in xrange(num_items):
+        item = batch.item.add()
+        item.id = item_id
+        field = item.field.add()
+        for token_id in range(0, num_tokens):
```
romovpa: just `range(num_tokens)`

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+            field.token_id.append(token_id)
+            background_count = random.randint(1, 5) if (token_id >= 40) else 0
+            topical_count = 10 if (token_id < 40) and ((token_id % 10) == (item_id % 10)) else 0
```
romovpa: denote the number of those topics with a correct word: 
meaningful_, reasonable_topics, target_topics, ...

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+        score_data.ParseFromString(results)
+        perplexity_score = messages.PerplexityScore()
+        perplexity_score.ParseFromString(score_data.data)
+        
+        if iter == 3:
+            assert abs(perplexity_score.value - 22.5) < 1.0
+        if iter == 8:
+            assert abs(perplexity_score.value - 21.7) < 1.0
+        
+        print "Iteration#" + str(iter) + ": Perplexity = %.3f" % perplexity_score.value
```
romovpa: get rid of the string concatenation

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+    score_data.ParseFromString(results)
+    top_tokens_score = messages.TopTokensScore()
+    top_tokens_score.ParseFromString(score_data.data)
+
+    print 'Top tokens per topic:',
+    topic_index = -1
+    for i in xrange(top_tokens_score.num_entries):
+        if top_tokens_score.topic_index[i] != topic_index:
+            topic_index = top_tokens_score.topic_index[i]
+            print "\nTopic#" + str(topic_index + 1) + ": ",
```
romovpa: string concatenation

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+        assert abs(top_tokens_score.weight[i] - 0.1) < 0.15
+        print top_tokens_score.token[i] + "(%.3f) " % top_tokens_score.weight[i],
```
romovpa: string concatenation

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+    print '\n',
+
+    shutil.rmtree(batches_folder)
```
romovpa: add newline at the end of file

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+
+    print 'Top tokens per topic:',
+    topic_index = -1
+    for i in xrange(top_tokens_score.num_entries):
+        if top_tokens_score.topic_index[i] != topic_index:
+            topic_index = top_tokens_score.topic_index[i]
+            print "\nTopic#" + str(topic_index + 1) + ": ",
+        assert abs(top_tokens_score.weight[i] - 0.1) < 0.15
+        print top_tokens_score.token[i] + "(%.3f) " % top_tokens_score.weight[i],
+    print '\n',
```
romovpa: Don't print incomplete lines (the lines without the '\n').
Instead, collect all the data needed to print the whole line and then print.

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+    ref_score_config = master_config.score_config.add()
+    ref_score_config.name = 'PerplexityScore'
+    ref_score_config.type = constants.ScoreConfig_Type_Perplexity
+    ref_score_config.config = messages.PerplexityScoreConfig().SerializeToString()
+
+    ref_score_config = master_config.score_config.add()
+    ref_score_config.name = 'TopTokensScore'
+    ref_score_config.type = constants.ScoreConfig_Type_TopTokens
+    config = messages.TopTokensScoreConfig()
+    config.num_tokens = 4
```
romovpa: name the constant at the beginning of the test

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+    ref_score_config.config = config.SerializeToString()
+
+    master_id = lib.ArtmCreateMasterComponent(master_config)
+
+    # Initialize model
+    init_args = messages.InitializeModelArgs()
+    init_args.model_name = pwt
+    init_args.disk_path = batches_folder
+    init_args.source_type = constants.InitializeModelArgs_SourceType_Batches
+    init_args.topics_count = 10
```
romovpa: constant

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+    lib.ArtmInitializeModel(master_id, init_args)
+
+    # Create configuration for batch processing
+    proc_args = messages.ProcessBatchesArgs()
+    proc_args.pwt_source_name = pwt
+    proc_args.nwt_target_name = nwt
+    for name in os.listdir(batches_folder):
+        proc_args.batch_filename.append(os.path.join(batches_folder, name))
+    proc_args.inner_iterations_count = 10
```
romovpa: constant

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+        lib.ArtmNormalizeModel(master_id, norm_args)    
+
+        # Retrieve and print perplexity score
+        results = lib.ArtmRequestScore(master_id, perplexity_args)
+        score_data = messages.ScoreData()
+        score_data.ParseFromString(results)
+        perplexity_score = messages.PerplexityScore()
+        perplexity_score.ParseFromString(score_data.data)
+        
+        if iter == 3:
```
romovpa: make the constants:
```
perplexity_tol = 1.0
expected_perplexity_value_on_iteration = {
    3: 22.5,
    8: 21.7,
}
```

python/wrapper_tests/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # Retrieve and print top tokens score
+    results = lib.ArtmRequestScore(master_id, top_tokens_args)
+    score_data = messages.ScoreData()
+    score_data.ParseFromString(results)
+    top_tokens_score = messages.TopTokensScore()
+    top_tokens_score.ParseFromString(score_data.data)
+
+    print 'Top tokens per topic:',
+    topic_index = -1
```
romovpa: In python you should not iterate lists through getting value with the index.
Instead, use `zip()`, `enumerate()` and the `itertools` module to make iteration concise. More ad
```python
import itertools

top_tokens_triplets = zip(top_tokens_score.topic_index, zip(top_tokens_score.token, top_tokens_score.weight))
for topic_index, group in itertools.groupby(top_tokens_triplets, key=lambda (topic_index, _): topic_index):
    for _, (token, weight) in group:
        ...
```

romovpa:
Travis-CI build fails because `python setup.py install` don't install submodules of the `artm` module. Maybe travis have an old version of `setuptools`.

Try to replace list of modules with [`find_packages()`](https://pythonhosted.org/setuptools/setuptools.html#id9) in `setup.py`:
```python
from setuptools import find_packages
setup(
    ...  
    packages=find_packages(),
    ...
)
```

or add them manually:
```python
packages=['artm', 'artm.wrapper']
```

python/setup.py
```diff
@@ -1,11 +1,10 @@
 from setuptools import setup
+from setuptools import find_packages
```
romovpa: `from setuptools import setup, find_packages` ;)

python/tests/wrapper/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,148 @@
...
+import artm.wrapper
+import artm.wrapper.messages_pb2 as messages
+import artm.wrapper.constants as constants
+import random
+import uuid
+import shutil
+import os
+import tempfile
+import itertools
+import pytest
```
romovpa: [PEP8: Imports](https://www.python.org/dev/peps/pep-0008/#id17)
Pay attention on how imports should be grouped and ordered

python/tests/wrapper/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,148 @@
...
+    print 'Top tokens per topic:'
+    top_tokens_triplets = zip(top_tokens_score.topic_index, zip(top_tokens_score.token, top_tokens_score.weight))
+    for topic_index, group in itertools.groupby(top_tokens_triplets, key=lambda (topic_index, _): topic_index):
+        string = 'Topic#{0} : '.format(topic_index)
+        for _, (token, weight) in group:
+            assert abs(weight - expected_top_tokens_weight) < top_tokens_tol
+            string += ' {0}({1:.3})'.format(token, weight)
+        print string
+
+    shutil.rmtree(batches_folder)
```
romovpa: Hm. This is dangerous because in the case of exception in the code above your temp folder will not be removed.
Here discussed a more robust way of working with temporary dir: [stackoverflow](http://stackoverflow.com/questions/13379742/right-way-to-clean-up-a-temporary-folder-in-python-class).

romovpa: pytest documentation describes the right way of working with temporary files: [Temporary directories and files](https://pytest.org/latest/tmpdir.html):
```python
def test_func(tmpdir):
    ...
   batches_folder = tmpdir.mkdir('batches') 
```

`tmpdir` is a [`py.path.local`](http://py.readthedocs.org/en/latest/path.html) object that casts to str as an absolute filename, e.g. on my machine:
```python
>>> print str(tmpdir)
/private/var/folders/3t/530dlyn937qc0k4ryxkcwz5wg8mc1p/T/pytest-1/test_func0
```

python/tests/wrapper/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # Retrieve and print perplexity score
+        results = lib.ArtmRequestScore(master_id, perplexity_args)
+        score_data = messages.ScoreData()
+        score_data.ParseFromString(results)
+        perplexity_score = messages.PerplexityScore()
+        perplexity_score.ParseFromString(score_data.data)
+
+        assert abs(perplexity_score.value - expected_perplexity_value_on_iteration[iter]) < perplexity_tol
+        print 'Iteration#{0} : Perplexity = {1:.3}'.format(str(iter), perplexity_score.value)
```
romovpa: format cast arguments to string by default:
```python
...format(iter, perplexity_score.value)
```

python/tests/wrapper/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,148 @@
...
+import itertools
+import pytest
+
+def test_func():
+    # Set some constants
+    num_tokens = 60
+    num_items = 100
+    batches_folder = tempfile.mkdtemp()
+    pwt = 'pwt'
+    nwt = 'nwt'
```
romovpa: Maybe you still need the constants in artm.wrapper.constants for this purpose?

python/tests/wrapper/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,148 @@
...
+    for token_id in xrange(num_tokens):
+        batch.token.append('token_{0}'.format(token_id))
+
+    for item_id in xrange(num_items):
+        item = batch.item.add()
+        item.id = item_id
+        field = item.field.add()
+        for token_id in xrange(num_tokens):
+            field.token_id.append(token_id)
+            background_count = random.randint(1, 5) if (token_id >= 40) else 0
```
romovpa: what about random in the tests?

python/tests/wrapper/test_01_synthetic_collection.py
MelLain: Nope, I think these constants are too local.

python/tests/wrapper/test_01_synthetic_collection.py
MelLain: The random is too weak and covered with asserts.

romovpa: Never use random in tests, it makes them stochastic.

Imagine the situation: you receive a build fail, tried to change the code to fix the potential problem but don't fix them actually, then you re-run tests and they are ok because of happiness of random number! You don't fix the problem, but CI system says you are.

At least you should set the random seed: `random.seed(constant_number)`. It is better to wrap your test in a parametrized function `test01(background_count)` and run this test for several fixed argument values.

python/tests/wrapper/test_02_regularizers.py
```diff
@@ -0,0 +1,133 @@
...
+        lib = artm.wrapper.LibArtm()
+        helper = helpers.TestHelper(lib)
+        
+        # Parse collection from disk
+        helper.parse_collection_uci(os.path.join(os.getcwd(), docword),
+                                    os.path.join(os.getcwd(), vocab),
+                                    batches_folder,
+                                    dictionary_name)
+
+        # Create master component and scores
```
sashafrey: Consider adding a method that takes maps score config to its type (PerplexityScoreConfig -> ScoreConfig_Type_Perplexity, etc). I'm not sure what would be the most pythonic way --- either a function with switch, or a map from type to the constant. But it would be great to just specify config, and let helper.create_master_component automatically figure out the constant for the type.

python/tests/wrapper/test_02_regularizers.py
```diff
@@ -0,0 +1,133 @@
...
+            helper.regularize_model(master_id, pwt, nwt, rwt,
+                                    ['SmoothSparsePhi', 'DecorrelatorPhi'], [smsp_phi_tau, decor_phi_tau])
+            helper.normalize_model(master_id, pwt, nwt, rwt)   
+
+            # Retrieve scores
+            perplexity_score = messages.PerplexityScore()
+            helper.retrieve_score(master_id, pwt, 'Perplexity', perplexity_score)
+
+            sparsity_phi_score = messages.SparsityPhiScore()
+            helper.retrieve_score(master_id, pwt, 'SparsityPhi', sparsity_phi_score)
```
sashafrey: I would prefer this syntax:
'''
sparsity_phi_score = helper.retrieve_score(master_id, pwt, 'SparsityPhi)
'''
Are there any downsides to this solution?

python/tests/wrapper/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,100 @@
...
+
+        # Create master component and scores
+        config = messages.TopTokensScoreConfig()
+        config.num_tokens = num_top_tokens
+        scores = [('PerplexityScore', constants.ScoreConfig_Type_Perplexity, messages.PerplexityScoreConfig()),
+                  ('TopTokensScore', constants.ScoreConfig_Type_TopTokens, config)]
+        master_id = helper.create_master_component(scores=scores)
+
+        # Initialize model
+        helper.initialize_model(master_id, pwt, num_topics, source_type='batches', disk_path=batches_folder)
```
sashafrey: master_id is a common parameter to many methods in helper. I suggest to make it optional (e.g. default to None), and add a setter helper.master_id = ...

python/tests/wrapper/test_01_synthetic_collection.py
```diff
@@ -0,0 +1,100 @@
...
+        print 'Top tokens per topic:'
+        top_tokens_triplets = zip(top_tokens_score.topic_index, zip(top_tokens_score.token, top_tokens_score.weight))
+        for topic_index, group in itertools.groupby(top_tokens_triplets, key=lambda (topic_index, _): topic_index):
+            string = 'Topic#{0} : '.format(topic_index)
+            for _, (token, weight) in group:
+                assert abs(weight - expected_top_tokens_weight) < top_tokens_tol
+                string += ' {0}({1:.3f})'.format(token, weight)
+            print string
+    finally:
+        shutil.rmtree(batches_folder)
```
sashafrey: new line at the end of file

python/tests/wrapper/test_05_multimodal.py
```diff
@@ -0,0 +1,215 @@
...
+
+        # Save batch and dictionary on the disk
+        lib.ArtmSaveBatch(batches_folder, batch)
+
+        dict_config.name = dictionary_name
+        with open(os.path.join(batches_folder, dictionary_name), 'wb') as file:
+            file.write(dict_config.SerializeToString())
+
+        # Create master component and scores
+        sp_phi_score_rus = messages.SparsityPhiScoreConfig()
```
sashafrey: Is there a pythonic way to do this in one line?
```
        sp_phi_score_rus = messages.SparsityPhiScoreConfig()
        sp_phi_score_rus.class_id = russian_class
```
In C# this can be done via initializers: `new SparsityPhiScoreConfig {class_id = russian_class}`. Ain't there something similar in python?

src/artm_tests/repeatable_result_test.cc
```diff
@@ -80,6 +81,12 @@ TEST(RepeatableResult, RandomGenerator) {
   }
 }
 
+TEST(RepeatableResult, TokenHasher) {
```
sashafrey: Remove this change (a similar change is done in https://github.com/bigartm/bigartm/pull/266)

PR #267, opened 2015-08-14
Java hash function as a string hash (when used as a seed)

sashafrey:


romovpa:
Why does it called "java" hash function?

sashafrey:
It comes from Java has string:
http://stackoverflow.com/questions/1660501/what-is-a-good-64bit-hash-function-in-java-for-textual-strings
Do you think this might be patented?

romovpa:
no, i'm just surprised )

bt2901:
I have some reservations regarding `31` constant. Since `char` stores values in `[0..255]` range, 31 is too small to separate low-level pieces of information. The number 31 is even more dangerous because the index of a capital letter usually differs from the index of the same lowercase letter by 32.

Assume that we care only about two-letter words, then the computation becomes

```h = 31 * (31 * h + token.keyword[0]) + token.keyword[1]```

It's very easy to produce an example of plausible-looking words hashing to the same value:

```
>>> h = 1125899906842597  # as in code, but any number will do
>>> func = lambda x, y: (31 * (31 * h + x) + y)
>>> func(ord('H'), ord('i'))
1081989810475738054
>>> func(ord('I'), ord('J'))
1081989810475738054
```
The collision becomes even more probable when dealing with longer words. This could lead to bad Phi initialization (and I think it indeed happens in some cases). Namely, the row corresponding to the word `Hi` will be always equal to the row corresponding to the word `IJ` after random initialization, whatever random seed one uses (assuming both are in the same modality). 


PS: I'm not sure how BigARTM deals with unicode characters, but I think that the same idea should hold there as well.

PR #272, opened 2015-09-07
ArtmAsyncRequestProcessBatches

sashafrey:
New API to enqueue ``ArtmRequestProcessBatches`` operation, and return immediately. The call returns an integer identifier of an ongoing operation. Use ``ArtmAwaitOperation`` method to check for the status of the operation.

For now this is just an experimental code. I'll split it into smaller commits, do further refactoring, include python wrappers, etc.


sashafrey:
@MelLain @AnyaP Could you please review changes in ``srcmain.cpp``, method ``ExecuteAsyncOnlineAlgorithm`` ? This is the core logic of new online async algorithm. Seems to work...

PR #283, opened 2015-09-13
fit_offline fixes

AnyaP:
- ARTM model has no longer num_document_passes field
- fit_offline has no longer reset_scores parameter, it is True forever
- fit_offline has no longer apply_weight, decay_weight parameters, they are redundant
- fit_offline has num_document_passes=1, reuse_theta=True parameters
- process_batches has reuse_theta=False parameter
- bug with dictionary_name for TopicKernelScore fixed

### DIFF
--- a/python/artm/model.py
+++ b/python/artm/model.py
@@ -40,8 +40,6 @@ class ARTM(object):
       class_ids (dict): list of class_ids and their weights to be used in model,
       key --- class_id, value --- weight, if not specified then all class_ids
       will be used
-      num_document_passes (int): number of iterations over each document
-      during processing, default=10
       cache_theta (bool): save or not the Theta matrix in model. Necessary
       if ARTM.get_theta() usage expects, default=True
       scores(list): list of scores (objects of artm.***Score classes), default=None
@@ -66,11 +64,10 @@ class ARTM(object):
 
     # ========== CONSTRUCTOR ==========
     def __init__(self, num_processors=0, topic_names=None, num_topics=10,
-                 class_ids=None, num_document_passes=10, cache_theta=True,
+                 class_ids=None, cache_theta=True,
                  scores=None, regularizers=None):
         self._num_processors = 0
         self._num_topics = 10
-        self._num_document_passes = 10
         self._cache_theta = True
 
         if topic_names is None or not topic_names:
@@ -88,9 +85,6 @@ class ARTM(object):
 
         if num_processors > 0:
             self._num_processors = num_processors
-
-        if num_document_passes > 0:
-            self._num_document_passes = num_document_passes
 
         if isinstance(cache_theta, bool):
             self._cache_theta = cache_theta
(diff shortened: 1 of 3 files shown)


MelLain:
Please, run codestyle_chechs.sh and fix codestyle problems before submitting any change

python/artm/model.py
```diff
@@ -66,11 +66,10 @@ class ARTM(object):
 
     # ========== CONSTRUCTOR ==========
     def __init__(self, num_processors=0, topic_names=None, num_topics=10,
```
MelLain: remove num_document_passes from constructor docstring

python/artm/model.py
```diff
@@ -250,21 +246,19 @@ def remove_dictionary(self, dictionary_name=None):
         else:
             raise IOError('dictionary_name is None')
 
-    def fit_offline(self, batch_vectorizer=None, num_collection_passes=1, decay_weight=0.0,
-                    apply_weight=1.0, reset_theta_scores=False):
+    def fit_offline(self, batch_vectorizer=None, num_collection_passes=10,
+                    num_document_passes=1, reuse_theta=True):
```
MelLain: What's the reason to set 10 as default value of num_collection_passes?

AnyaP: If a user wants to "fit_offline" his model and he doesn't want to think about anything else, how many passes should be performed? For sure, it depends on the size of the collection, but fit_offline is supposed to be performed for small collections like kos or NIPS or MRRO. Then, 1 is not enough, while too many might take too long. So 10 =)

python/artm/model.py
```diff
@@ -398,7 +386,6 @@ def fit_online(self, batch_vectorizer=None, tau0=1024.0, kappa=0.7,
                                             nwt='nwt_hat',
                                             regularizer_name=theta_reg_name,
                                             regularizer_tau=theta_reg_tau,
-                                            num_inner_iterations=self._num_document_passes,
```
MelLain: Maybe I don't understand... but where the num_document_passes parameter in fit_online()? I think, you need to add it in the method parameters, in the master_component.process_batches() call, and describe it in the docstring.

python/artm/scores.py
```diff
@@ -490,8 +490,8 @@ def __init__(self, name=None, class_id=None, topic_names=None, eps=None,
 
         self._dictionary_name = ''
         if dictionary_name is not None:
-            self._self._dictionary_name = dictionary_name
-            config.cooccurrence_dictionary_name = dictionary_name
+            self._dictionary_name = dictionary_name
```
MelLain: So stupid mistake... Thanks)

python/artm/model.py
AnyaP: Done.

python/artm/model.py
AnyaP: You are right, fixed.

sashafrey:
:1
@MelLain Merge in?

PR #286, opened 2015-09-15
Ptdw prototype

sashafrey:
This change makes a cleanup and refactoring in the code of ``processor.cc`` to exposes a ``p_tdw`` matrix. This matrix is essential for experiments with alternative text representation beyond simple bag-of-words format. Currently we are able to consume this information in Vowpal Wabbit format, where each document is represented with a sequence of words. BigARTM will preserve this information in ``Batch.Token.Field.token_id`` field. The code of processor.cc can benefit from this, and implement certain operations on ``p_tdw`` matrix.
The next steps are to 
* experiment with ``p_tdw`` matrix directly in ``processor.cc`` code ( @anyap )
* implement a feature to retrieve ``p_tdw`` matrix via ``ThetaMatrix`` message ( @MelLain )

### DIFF
--- a/python/artm/messages_pb2.py
+++ b/python/artm/messages_pb2.py
@@ -13,7 +13,7 @@ from google.protobuf import descriptor_pb2
 DESCRIPTOR = _descriptor.FileDescriptor(
   name='artm/messages.proto',
   package='artm',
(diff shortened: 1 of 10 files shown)


sashafrey:
I've measured performance of the new method, and it is surprisingly good. In fact it is even slightly better than the current version. At least on my machine with 4 cores.

sashafrey:
Approved by @MelLain

PR #292, opened 2015-09-16
Mel lain different

MelLain:
- assymetric cooc values support
- coherence regularizer speed up

### DIFF
--- a/src/artm/core/dictionary.cc
+++ b/src/artm/core/dictionary.cc
@@ -36,18 +36,20 @@ Dictionary::Dictionary(const artm::DictionaryConfig& config) {
           first_cooc_iter = cooc_values_.find(first_index_iter->second);
         }
 
-        auto second_cooc_iter = cooc_values_.find(second_index_iter->second);
-        if (second_cooc_iter == cooc_values_.end()) {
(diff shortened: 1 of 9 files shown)


src/artm/core/dictionary.cc
```diff
@@ -43,11 +43,13 @@ Dictionary::Dictionary(const artm::DictionaryConfig& config) {
...
-        // the data representation is symmetric for both first and second tokens in cooc pair
         first_cooc_iter->second.insert(std::make_pair(second_index_iter->second,
                                                       config.cooc_entries().value(i)));
-        second_cooc_iter->second.insert(std::make_pair(first_index_iter->second,
-                                                       config.cooc_entries().value(i)));
+
+        if (config.cooc_entries().symmetric_cooc_values()) {
+          second_cooc_iter->second.insert(std::make_pair(first_index_iter->second,
+                                                         config.cooc_entries().value(i)));
+        }
```
AnyaP: Can we place strings 39-43 also under if(config.cooc_entries().symmetric_cooc_values())? 
To create second_cooc_iter only if we are going to use it.

PR #293, opened 2015-09-18
New score and small code cleanup

MelLain:
- update parameters preparing process in scores (use the similar way as in regularizers)
- add new TopicPartInNwt score (THE NAME IS TEMPORARY :) )

### DIFF
--- a/src/artm/messages.proto
+++ b/src/artm/messages.proto
@@ -240,6 +240,7 @@ message ScoreConfig {
     TopTokens = 4;
     ThetaSnippet = 5;
     TopicKernel = 6;
+    TopicMassPhi = 7;
   }
 
   optional string name = 1;
@@ -257,6 +258,7 @@ message ScoreData {
     TopTokens = 4;
     ThetaSnippet = 5;
     TopicKernel = 6;
+    TopicMassPhi = 7;
   }
 
   optional string name = 1;
@@ -385,6 +387,20 @@ message TopicKernelScore {
   optional float average_coherence = 8;
   repeated StringArray kernel_tokens = 9;
   optional StringArray topic_name = 10;
+}
+
+// Represents a configuration of a topic part in nwt score
+message TopicMassPhiScoreConfig {
+  optional float eps = 1 [default = 1e-37];
+  optional string class_id = 2 [default = "@default_class"];
+  repeated string topic_name = 3;
+}
+
+// Represents a result of calculation of a topic part in nwt score
+message TopicMassPhiScore {
+  optional double value = 1;
+  repeated string topic_name = 2;
+  repeated double topic_mass = 3;
 }
 
 // Represents a topic model (or an operation on topic model)
--- a/src/artm/score/perplexity.cc
+++ b/src/artm/score/perplexity.cc
@@ -6,6 +6,7 @@
 #include <sstream>
 
 #include "artm/core/exceptions.h"
+#include "artm/core/protobuf_helpers.h"
 
 #include "artm/score/perplexity.h"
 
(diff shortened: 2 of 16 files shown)


sashafrey:
@anyap Any thoughts about the name for this score?
My prefered version is `TopicMassPhi`.
- It should be `Phi` and not `Nwt`, because some people may want to use this score for regularization matrix `Rwt`.
- In the future we may implement a similar score `TopicMassTheta`, based on theta matrix. I like that the names follow the pattern set up by `SparsityPhi` and `SparsityTheta`.
- I'm not sure about `Mass`, but I've spent some time in http://www.thesaurus.com/ and couldn't figure out a better concept. I've been thinking about `Size`, `Weight`, `Volume`, but to me `Mass` is more reasonable.

src/artm/messages.proto
```diff
@@ -387,6 +389,21 @@ message TopicKernelScore {
   optional StringArray topic_name = 10;
 }
 
+// Represents a configuration of a topic part in nwt score
+message TopicPartInNwtScoreConfig {
+  optional float eps = 1 [default = 1e-37];
+  optional string class_id = 2 [default = "@default_class"];
+  repeated string topic_name = 3;
+  repeated string topic_name_to_score = 4;
```
sashafrey: I think it is getting a little bit too complicated with two lists (`topic_name` and `topic_name_to_score`). Maybe it would be sufficient to only keep one list (`topic_name`), and calculate the cumulative mass across `topic_name` devided by cumulative mass of **all topics**? In addition the score should output a vector of cumulative mass for each topic. This will enable the user to calculate the ratio between any two sets of topics.

The resulting messages would then look as follows:
```
message TopicMassPhiScoreConfig {
  optional float eps = 1 [default = 1e-37];
  optional string class_id = 2 [default = "@default_class"];
  repeated string topic_name = 3;
}

message TopicMassPhiScore {
  optional double value = 1;
  repeated string topic_name = 2; 
  repeated double topic_mass = 3;
}
```

The `TopicMassPhiScore.topic_mass ` will give the cumulative score for each topic. The names of each topic will be in `TopicMassPhiScore.topic_name`.  Note that the field `TopicMassPhiScore.topic_name` will contains all topics, e.g. it will not match the content of the input field `TopicMassPhiScoreConfig.topic_name`. I think this is OK.

src/artm/score/perplexity.cc
```diff
@@ -29,34 +30,20 @@ void Perplexity::AppendScore(
...
-
-    for (int topic_id = 0; topic_id < config_.theta_sparsity_topic_name_size(); ++topic_id) {
-      for (int real_topic_id = 0; real_topic_id < topics_size; ++real_topic_id) {
-        if (topic_name.Get(real_topic_id) == config_.theta_sparsity_topic_name(topic_id)) {
-          topics_to_score[real_topic_id] = true;
-          topics_to_score_size++;
-          break;
-        }
-      }
-    }
```
sashafrey: Thanks for cleaning this up!

src/artm/messages.proto
MelLain: Well. the first note is OK. About the second one.. it seems to me, that current field topic_part in score result contains values, you'd like to see in repeated double topic_mass = 3, the only change is to rename the field)

18.09.2015, 18:50, "sashafrey" <notifications@github.com>:

sashafrey: Almost... I think ``topic_mass(i)`` should contain ``topic_part(i)`` instead of ``topic_part(i) / denominator``.

sashafrey: @MelLain , do you think ``TopicMassPhi`` would be a good name? I think @anyap was ok with it, so if you too agree then let's conclude on it :)

PR #300, opened 2015-09-20
Ptwd extraction

MelLain:
Add ability to cache and to extract the Ptdw matrix.


src/artm/core/processor.cc
```diff
@@ -41,6 +41,77 @@ const float kProcessorEps = 1e-16;
 namespace artm {
 namespace core {
 
+static void CreateThetaCacheEntry(std::shared_ptr<DataLoaderCacheEntry> new_cache_entry_ptr,
+                                  DenseMatrix<float>* theta_matrix,
+                                  const Batch& batch,
+                                  const ModelName& model_name,
+                                  std::shared_ptr<ModelIncrement> model_increment,
+                                  int topic_size) {
+  if (new_cache_entry_ptr != nullptr) {
```
sashafrey: Reduce nesting:
```
if (new_cache_entry_ptr == nullptr)
  return;
```

src/artm/core/processor.cc
```diff
@@ -41,6 +41,77 @@ const float kProcessorEps = 1e-16;
...
+                                 const ModelName& model_name,
+                                 std::shared_ptr<ModelIncrement> model_increment,
+                                 int item_id,
+                                 int topic_size) {
+  if (new_cache_entry_ptr != nullptr) {
+    new_cache_entry_ptr->set_batch_uuid(batch_uuid);
+    new_cache_entry_ptr->set_model_name(model_name);
+    new_cache_entry_ptr->mutable_topic_name()->CopyFrom(model_increment->topic_model().topic_name());
+    new_cache_entry_ptr->add_item_title(std::string());
+    std::cout << "Ptwd: " << ptdw_matrix->no_rows() << ' ' << ptdw_matrix->no_columns() << std::endl;
```
sashafrey: Don't forget to remove std::cout :)

src/artm/core/processor.cc
```diff
@@ -41,6 +41,77 @@ const float kProcessorEps = 1e-16;
...
+    std::cout << "Modl: " << model_increment->topic_model().token_size() << ' ' << topic_size << std::endl;
+
+    for (int token_index = 0; token_index < ptdw_matrix->no_rows(); ++token_index) {
+      new_cache_entry_ptr->add_item_id(item_id);
+      auto non_zero_topic_values = new_cache_entry_ptr->add_theta();
+      auto non_zero_topic_indices = new_cache_entry_ptr->add_topic_index();
+      for (int topic_index = 0; topic_index < topic_size; ++topic_index) {
+        float value = ptdw_matrix->operator()(token_index, topic_index);
+        if (std::fabs(value) < kProcessorEps) {
```
sashafrey: ``>`` ?

src/artm/core/processor.cc
```diff
@@ -41,6 +41,77 @@ const float kProcessorEps = 1e-16;
...
+
+static void CreatePtdwCacheEntry(std::shared_ptr<DataLoaderCacheEntry> new_cache_entry_ptr,
+                                 DenseMatrix<float>* ptdw_matrix,
+                                 const std::string& batch_uuid,
+                                 const ModelName& model_name,
+                                 std::shared_ptr<ModelIncrement> model_increment,
+                                 int item_id,
+                                 int topic_size) {
+  if (new_cache_entry_ptr != nullptr) {
+    new_cache_entry_ptr->set_batch_uuid(batch_uuid);
```
sashafrey: I would rather move the following lines out from this method, and keep them in the ``ThreadFunction``.
```
new_cache_entry_ptr->set_batch_uuid(batch_uuid);
new_cache_entry_ptr->set_model_name(model_name);
new_cache_entry_ptr->mutable_topic_name()->CopyFrom(model_increment->topic_model().topic_name());
```
This would reduce the number of input parameters for ``CreatePtdwCacheEntry`` (and same for ``CreateThetaCacheEntry``. Also, copying ``topic_name`` for all items might be an overhead.

src/artm/core/processor.cc
```diff
@@ -467,13 +541,19 @@ InferThetaAndUpdateNwtSparse(const ModelConfig& model_config, const Batch& batch
...
+  CreateThetaCacheEntry(new_cache_entry_ptr, theta_matrix, batch, part->model_name(), model_increment, topics_count);
 }
 
 static void
 InferPtdwAndUpdateNwtSparse(const ModelConfig& model_config, const Batch& batch, float batch_weight, const Mask* mask,
                             const InstanceSchema& schema, const CsrMatrix<float>& sparse_ndw,
                             const ::artm::core::PhiMatrix& p_wt, DenseMatrix<float>* theta_matrix,
-                            NwtWriteAdapter* nwt_writer, util::Blas* blas) {
+                            NwtWriteAdapter* nwt_writer, util::Blas* blas,
+                            std::shared_ptr<DataLoaderCacheEntry> new_cache_entry_ptr = nullptr,
```
sashafrey: Using shared pointer is a bad practice. Its better to replace it with native pointer. This is because caller can always get a native pointer from ``std::shared_ptr``, but not the other way around. There are cases when passing ``std::shared_ptr`` is appropriate, but they should be related to some kind of memory management (like passing an object to and from ProcessorQueue in BigARTM).

src/artm/core/master_component.cc
```diff
@@ -350,7 +356,8 @@ void MasterComponent::RequestProcessBatches(const ProcessBatchesArgs& process_ba
     auto pi = std::make_shared<ProcessorInput>();
     pi->set_notifiable(&batch_manager);
     pi->set_scores_merger(scores_merger);
-    pi->set_cache_manager(cache_manager_ptr);
+    pi->set_cache_manager(theta_cache_manager_ptr);
+    pi->set_ptdw_cache_manager(ptdw_cache_manager_ptr);
     pi->set_model_name(model_name);
     pi->set_batch_filename(args.batch_filename(batch_index));
     pi->set_batch_weight(args.batch_weight(batch_index));
```
sashafrey: Eventually we should be able to completely remove ``ProcessBatchesArgs.use_ptdw_matrix``. For now I think we should add the following logic:
```
if (process_batches_args.theta_matrix_type == Ptdw)
  pi->mutable_model_config()->set_use_ptdw_matrix(true)
```

src/artm/core/processor.cc
```diff
@@ -469,13 +537,19 @@ InferThetaAndUpdateNwtSparse(const ModelConfig& model_config, const Batch& batch
...
 
 static void
 InferPtdwAndUpdateNwtSparse(const ModelConfig& model_config, const Batch& batch, float batch_weight, const Mask* mask,
                             const InstanceSchema& schema, const CsrMatrix<float>& sparse_ndw,
                             const ::artm::core::PhiMatrix& p_wt, DenseMatrix<float>* theta_matrix,
-                            NwtWriteAdapter* nwt_writer, util::Blas* blas) {
+                            NwtWriteAdapter* nwt_writer, util::Blas* blas,
+                            DataLoaderCacheEntry* new_cache_entry_ptr = nullptr,
+                            DataLoaderCacheEntry* new_ptdw_cache_entry_ptr = nullptr,
+                            ModelIncrement* model_increment = nullptr,
```
sashafrey: Would it be possible to avoid passing model_increment and part? I think they are spreading across the code for a  very minor reason --- you use them for in CreateThetaCacheEntry to do this:

+  new_cache_entry_ptr->set_batch_uuid(batch.id());
+  new_cache_entry_ptr->set_model_name(model_name);
+  new_cache_entry_ptr->mutable_topic_name()->CopyFrom(model_increment->topic_model().topic_name());

I would rather move this back into ThreadFunction...

MelLain: Done!

22.09.2015, 23:51, "sashafrey" <notifications@github.com>:

sashafrey:
This pull request is overtaken by https://github.com/bigartm/bigartm/pull/305

PR #315, opened 2015-10-04
Implement framework for ptdw regularizers

sashafrey:
This pull request
* implements a framework for ``ptdw`` regularizers, and 
* implements first ``ptdw`` regularizer based on the code from https://github.com/bigartm/bigartm/compare/master...AnyaP:master
* removes ``use_ptdw_matrix`` flag. Now ``ptdw`` matrix will be used automatically if there is at least one ``ptdw`` regularizer, or if user performs ``ptdw`` matrix retrieval.

From user perspective ``ptdw`` regularizers are no different from ``theta`` regularizers. To apply ``ptdw`` regularizers user need to add their names and weights into ``ProcessBatchesArgs.regularizer_name`` and ``ProcessBatchesArgs.regularizer_tau`` lists.

This pull request does not update python APIs. @MelLain , could you please wrap ``SmoothPtdw`` into the corresponding class for Python interface? (eventually I also need to learn how to do this)

### DIFF
--- a/src/artm/messages.proto
+++ b/src/artm/messages.proto
@@ -116,7 +116,6 @@ message ModelConfig {
   optional bool use_new_tokens = 16 [default = true];
   optional bool opt_for_avx = 17 [default = true];
   repeated RegularizerSettings regularizer_settings = 18;
-  optional bool use_ptdw_matrix = 19 [default = false];
 }
 
 // Represents a configuration of a general regularizer
@@ -129,6 +128,7 @@ message RegularizerConfig {
     LabelRegularizationPhi = 4;
     SpecifiedSparsePhi = 5;
     ImproveCoherencePhi = 6;
+    SmoothPtdw = 7;
   }
 
   optional string name = 1;
@@ -183,6 +183,18 @@ message ImproveCoherencePhiConfig {
   repeated string topic_name = 1;
   repeated string class_id = 2;
   optional string dictionary_name = 3;
+}
+
+// Represents a configuration of a Smooth Ptdw regularizer
+message SmoothPtdwConfig {
+  enum Type {
+    MovingAverage = 1;
+    MovingProduct = 2;
+  }
+
+  optional Type type = 1 [default = MovingAverage];
+  optional int32 window = 3 [default = 10];
+  optional double threshold = 4 [default = 1.0];
 }
 
 // Represents an internal state of a general regularizer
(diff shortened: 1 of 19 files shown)


src/artm/messages.proto
```diff
@@ -185,6 +185,18 @@ message ImproveCoherencePhiConfig {
...
+// Represents a configuration of a Smooth Ptdw regularizer
+message SmoothPtdwConfig {
+  enum Type {
+    MovingAverage = 1;
+    MovingProduct = 2;
+  }
+
+  optional Type type = 1 [default = MovingAverage];
+  optional int32 window = 3 [default = 10];
+  optional double treshold = 4 [default = 1.0];
```
MelLain: maybe 'threshold'?)

src/artm/regularizer/smooth_ptdw.cc
```diff
@@ -0,0 +1,108 @@
+// Copyright 2014, Additive Regularization of Topic Models.
+
+// Author: Murat Apishev (great-mel@yandex.ru)
```
MelLain: Please, remove my name from here and .h file,it's not my code

src/artm/regularizer/smooth_ptdw.cc
```diff
@@ -0,0 +1,108 @@
...
+  if (config_.type() == SmoothPtdwConfig_Type_MovingAverage) {
+    // 1. evaluate wich tokens are background
+    double treshold = config_.treshold();
+    std::vector<bool> is_background(local_token_size, false);
+    int count_background = 0;
+    for (int i = 0; i < local_token_size; ++i) {
+      const float* local_ptdw_ptr = &(*ptdw)(i, 0);  // NOLINT
+      double sum_background = 0.0;
+      for (int k = 0; k < topics_count; ++k) {
+        char b = 'b';
```
MelLain: I don't understand this: why the first letter of background topic names is hard-coded?

src/artm/messages.proto
sashafrey: thanks! Fixed.

src/artm/regularizer/smooth_ptdw.cc
sashafrey: I replaced your name with @AnyaP  - this is her code :)

src/artm/regularizer/smooth_ptdw.cc
sashafrey: We definitely should address this issue, but perhaps later... Ideally we should describe which topics are background in parameters of this regularizer. However none of other regularizers has this concept. Also, what if user did not configure background topics? Does it mean that ``smooth_ptdw`` regularizer is not applicable?

Maybe an alternative solution would be to define a list of all background words as a dictionary, and ask user to build a separate topic model in order to create this dictionary.

PR #317, opened 2015-10-06
remove batches random naming + add new Ptdw reg in ARTM

MelLain:


python/artm/regularizers.py
```diff
@@ -115,18 +104,17 @@ def config(self):
     def type(self):
         return self._type
 
-    @topic_names.setter
-    def topic_names(self, topic_names):
-        _reconfigure_field(self, topic_names, 'topic_names')
+    @config.setter
```
sashafrey: ``@config.setter`` should invoke ``master_reconfigure()``, to be consistent with other setters

python/artm/wrapper/constants.py
```diff
@@ -12,6 +12,7 @@
 RegularizerConfig_Type_LabelRegularizationPhi = 4
 RegularizerConfig_Type_SpecifiedSparsePhi = 5
 RegularizerConfig_Type_ImproveCoherencePhi = 6
+RegularizerConfig_Type_SmoothPtdw = 7
 ScoreConfig_Type_Perplexity = 0
```
sashafrey: Please add these two constants:
```
  SmoothPtdwConfig_Type_MovingAverage = 1
  SmoothPtdwConfig_Type_MovingProduct = 2
```

PR #323, opened 2015-10-19
speed up coherence regularizer and counter

MelLain:


src/artm/regularizer/improve_coherence_phi.cc
```diff
@@ -41,23 +41,31 @@ bool ImproveCoherencePhi::RegularizePhi(const ::artm::core::PhiMatrix& p_wt,
     return false;
   }
 
+  // create the conversion from index if token in Dictionary -> index of token in Phi
+  std::vector<int> dict_to_phi_indices(dictionary_ptr->size(), 0);
+  for (auto& index_token : dictionary_ptr->index_token())
+    dict_to_phi_indices[index_token.first] = n_wt.token_index(index_token.second);
```
sashafrey: token_index might return -1 if token is not present in PhiMatrix. Check for this later when using dict_to_phi_indice.

PR #340, opened 2015-10-31
Add ability to use not only log function in KL-div in regularizers

MelLain:


sashafrey:
@MelLain Everything looks good (except few minor changes that I've mention in comments).

PR #342, opened 2015-11-01
Include c_interface in artm static

sashafrey:


sashafrey:
@JeanPaulShapo I tend to agree with the option that you suggested. 

* On Windows we have to include `c_interface.cc` when building shared library `artm.dll`, as we do today. Otherwise this DLL will have only private code (not callable from outside of the dll). On Windows it is required to add `__declspec(dllexport)` for everything callable outside of the dll. On Linux by default all content of the `.so` file is public.
* From what I see it does not harm to include `c_interface.cc` twice - first into `artm-static` static library, and then into shared library.

JeanPaulShapo:
@sashafrey This change looks OK

PR #339, opened 2015-10-30
Enable running bigartm-cli as standalone executable.

JeanPaulShapo:
It works on Linux (Fedora 21) without Intel MKL.
Before merging it's strongly recommended to test on Windows/Linux with/without Intel MKL
(i don't have Windows and Intel MKL binaries :( )

PR #343, opened 2015-11-02
Enable running bigartm-cli as standalone executable

sashafrey:


sashafrey:
Fixed in pull request #339

PR #358, opened 2015-11-19
MelLain dictionaries

MelLain:


sashafrey:
@MelLain 
I'm still reviewing the change, but I would like to suggest few things we can do to improve user upgrade experience:
- In ArtmImportDictionary verify that dictionary file has an extention .dict. If filename does not end with .dict we should throw an error, explaining that "Since BigARTM v0.7.4 function ArtmImportDictionary no longer accepts dictionaries create prior to v0.7.4. Dictionaries should be recreated via ArtmGatherDictionary method, and exported with .dict extension". Yes, this is unreliable, and users might already have a .dict file, or they may choose to rename file. However, this is at least something we can do to avoid a crash. Few releases later we will remove this check and let users specify any name for their dictionaries.  
-  do not delete methods ArtmRequestParseCollection, ArtmReconfigureDictionary and ArtmRequestLoadDictionary methods but throw an exception from them, explaining what users needs to do in order to upgrade their scripts.
- restore all deleted fields in CollectionParserConfig and InitializeModelArgs. In Helper::Validate(...) methods check whether those entries are empty; if they are not empty throw an exception that explains what user needs to do.  Mark the fields as "obsoleted" in messages.proto, so that we remember to delete them in the future.

src/artm/c_interface.cc
```diff
@@ -485,6 +485,8 @@ int ArtmOverwriteTopicModel(int master_id, int length, const char* topic_model)
 }
 
 int ArtmInitializeModel(int master_id, int length, const char* init_model_args) {
+  // check if given dictionary name is appear in instance.dictionaries_
```
sashafrey: Remove the comment. Methods of c_interface should only verify local consistency of protobuf messages. All other high-level preconditions should be verified by individual methods. Particularly, MasterComponetn::InitializeModel (or whatever method implements the initialization) needs to throw an error if given name is not present in the instance.dictionaries_.

src/artm/c_interface.cc
```diff
@@ -616,43 +646,17 @@ int ArtmDisposeBatches(int master_id, int length, const char* dispose_batches_ar
...
-    return last_message()->size();
-  } CATCH_EXCEPTIONS;
-}
-
 int ArtmParseCollection(int length, const char* collection_parser_config) {
   try {
     EnableLogging();
     artm::CollectionParserConfig config;
     ParseFromArray(collection_parser_config, length, &config);
-    ::artm::core::Helpers::FixAndValidate(&config, /* throw_error =*/ true);
```
sashafrey: Are you intentionally removing FixAndValidate for collection parser config?

utils/cpplint_files.txt
```diff
@@ -60,6 +60,7 @@ src/artm/core/data_loader.h
 src/artm/core/dense_phi_matrix.h
 src/artm/core/dictionary.h
 src/artm/core/exceptions.h
+src/artm/core/fileread_helpers.h
```
sashafrey: Could you move this file under `src/utility` and rename to `ifstream_or_cin.h`? This would discourage people from putting more stuff into this file, and generally this is a good practice...

src/artm/core/instance.cc
```diff
@@ -381,6 +398,18 @@ void Instance::DisposeDictionary(const std::string& name) {
   dictionaries_.erase(name);
 }
 
+void Instance::DisposeDictionaryImpl(const std::string& name) {
```
sashafrey: I think the code would be cleaner if you remove this method and move `dictionaries_impl_.erase(name)` into `Instance::DisposeDictionary` method. This would be similar to how merger handles old- and new-style models in `Merger::DisposeModel` method.
``` 
  topic_model_.erase(model_name);
  phi_matrix_.erase(model_name);
```

src/artm/core/instance.cc
```diff
@@ -372,6 +384,11 @@ void Instance::DisposeRegularizer(const std::string& name) {
   schema_.set(new_schema);
 }
 
+void Instance::CreateOrReconfigureDictionaryImpl(const DictionaryData& data) {
```
sashafrey: This method (`CreateOrReconfigureDictionaryImpl`) should create dictionary in two places -- both `dictionaries_impl_` and `dictionaries_`. Here is the right place to call a method that creates an old dictionary (`Dictionary`) from `DictionaryData`. This is a temporary hack that will ensure that all scores and regularizers still work if user creates a new dictionary.

src/artm/core/master_component.cc
```diff
@@ -227,9 +357,335 @@ void MasterComponent::AttachModel(const AttachModelArgs& args, int address_lengt
...
+    dictionary_data->add_class_id(token.class_id);
+    dictionary_data->add_token_df(entry.token_df());
+    dictionary_data->add_token_tf(entry.token_tf());
+    dictionary_data->add_token_value(entry.token_value());
+
+    old_index_new_index.insert(std::pair<int, int>(dictionary_token_index.find(token)->second,
+                                                   accepted_tokens_count - 1));
+  }
+
+  auto cooc_dictionary_data = std::make_shared<artm::DictionaryData>();
```
sashafrey: Wow.. This approach require you to implement filtering logic twice. I meant something simpler.. Just create a constructor of `::artm::core::Dictionary` class that accepts `::artm::DictionaryData` as a parameter. Then, use this constructor in `CreateOrReconfigureDictionaryImpl` to automatically create an old dictionary (every time the method is called).

src/artm/core/master_component.h
```diff
@@ -29,6 +29,15 @@ class TopicModel;
 class Score;
 class BatchManager;
 
+class TokenInfo {
```
sashafrey: Move `TokenInfo` into `master_component.cc`.

src/artm_tests/collection_parser_test.cc
```diff
@@ -103,9 +63,6 @@ TEST(CollectionParser, ErrorHandling) {
 
   config.set_vocab_file_path("../../../test_data/vocab.parser_test.txt");
   config.set_docword_file_path("no_such_file");
-  ASSERT_THROW(::artm::ParseCollection(config), artm::DiskReadException);
```
sashafrey: Remove last two lines from the test:
```
   config.set_vocab_file_path("../../../test_data/vocab.parser_test.txt");
   config.set_docword_file_path("no_such_file");
``

File: src/artm_tests/collection_parser_test.cc [hunk 9]
```diff
@@ -19,51 +19,11 @@ TEST(CollectionParser, UciBagOfWords) {
...
-  config.set_dictionary_file_name("test_parser.dictionary");
-  config.set_gather_cooc(true);
-  config.add_cooccurrence_token("token1");
-  config.add_cooccurrence_token("token2");
-  config.add_cooccurrence_token("token3");
   config.set_num_items_per_batch(1);
   config.set_vocab_file_path("../../../test_data/vocab.parser_test.txt");
   config.set_docword_file_path("../../../test_data/docword.parser_test.txt");
 
-  std::shared_ptr< ::artm::DictionaryConfig> dictionary_parsed = ::artm::ParseCollection(config);
```
sashafrey: Could you turn this test into a test of ArtmGatherDictionary? To do this you also need a method `ArtmRequestDictionary` that would return a `DictionaryData` message. I think it would be greate to do this as part of this pull request.

src/artm_tests/collection_parser_test.cc
```diff
@@ -132,19 +101,28 @@ TEST(CollectionParser, Multiclass) {
   ::artm::CollectionParserConfig config;
   config.set_format(::artm::CollectionParserConfig_Format_BagOfWordsUci);
   config.set_target_folder(target_folder);
-  config.set_dictionary_file_name("test_parser.dictionary");
   config.set_vocab_file_path("../../../test_data/vocab.parser_test_multiclass.txt");
   config.set_docword_file_path("../../../test_data/docword.parser_test.txt");
 
-  std::shared_ptr< ::artm::DictionaryConfig> dictionary_parsed = ::artm::ParseCollection(config);
```
sashafrey: Same question here as for the test above (` TEST(CollectionParser, UciBagOfWords)`).
Let's verify that `ArtmGatherDictionary` is able to handle multi-class vocab files.

sashafrey:
I merged the last commit (https://github.com/MelLain/bigartm/commit/4e2fcc2fee21b20c7cb33a2e9dca3bc885d4ea5e) in a separate pull request. The rest of this change is copied into a separate pull request (https://github.com/bigartm/bigartm/pull/377). To continue, please re-create your branch from https://github.com/sashafrey/bigartm/tree/MelLain-Dictionaries . I'll try to fix some of my comments in that branch.

utils/cpplint_files.txt
sashafrey: Done (https://github.com/sashafrey/bigartm/commit/2cc9264e8d6ce1aab9848c56fa170b2cadf00398)

src/artm/c_interface.cc
sashafrey: Done

src/artm/core/instance.cc
sashafrey: Done

src/artm/c_interface.cc
sashafrey: Yes, done

src/artm/core/instance.cc
sashafrey: Done

src/artm/core/master_component.cc
sashafrey: Done

src/artm_tests/collection_parser_test.cc
sashafrey: Done

PR #361, opened 2015-11-25
Excludes static linkage for Mac OS

JeanPaulShapo:
There are many discussions related to static linkage for Mac OS
https://stackoverflow.com/questions/3801011/ld-library-not-found-for-lcrt0-o-on-osx-10-6-with-gcc-clang-static-flag
https://stackoverflow.com/questions/5259249/creating-static-mac-os-x-c-build
It seems that nowadays we cannot build any static executable on Mac OS using GCC.
So, for Mac OS users option "BUILD_STATIC_BIGARTM" is unnecessary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,11 @@ else()
   message("-- Build type: ${CMAKE_BUILD_TYPE}")
 endif()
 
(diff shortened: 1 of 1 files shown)


sashafrey:
:+1:

PR #364, opened 2015-11-26
eliminating -Wsign-compare warnings

omtcyf0:
This patch aims to eliminate most of the `int` <-> `unsigned` comparisons used in `for` loops.

The CMake build system of the project is configured in a way `-Wall` is enabled and hence I expect the warnings to be informative and to show real issues connected with the written code, eliminating "junk" warnings (like described comparisons) seems quite reasonable for me.

Even though Google Code Style [is against using `unsigned` type](https://google.github.io/styleguide/cppguide.html#Integer_Types), it is widely used in many Google's projects. IMO there's nothing bad in using `unsigned`.

Plus, two warnings generated for `case` blocks not handling every item of given `enum` were eliminated via adding `default:`.

### DIFF
--- a/src/artm/core/collection_parser.cc
+++ b/src/artm/core/collection_parser.cc
@@ -574,7 +574,7 @@ std::shared_ptr<DictionaryConfig> CollectionParser::ParseVowpalWabbit() {
     std::string item_title = strs[0];
 
     ClassId class_id = DefaultClass;
-    for (int elem_index = 1; elem_index < strs.size(); ++elem_index) {
+    for (unsigned elem_index = 1; elem_index < strs.size(); ++elem_index) {
       std::string elem = strs[elem_index];
       if (elem.size() == 0)
         continue;
--- a/src/artm/core/dense_phi_matrix.cc
+++ b/src/artm/core/dense_phi_matrix.cc
@@ -84,7 +84,7 @@ PhiMatrixFrame::PhiMatrixFrame(const PhiMatrixFrame& rhs)
       token_collection_(rhs.token_collection_),
       spin_locks_() {
   spin_locks_.reserve(rhs.spin_locks_.size());
-  for (int i = 0; i < rhs.spin_locks_.size(); ++i)
+  for (unsigned i = 0; i < rhs.spin_locks_.size(); ++i)
     spin_locks_.push_back(std::make_shared<SpinLock>());
 }
 
--- a/src/artm/core/helpers.cc
+++ b/src/artm/core/helpers.cc
@@ -846,13 +846,13 @@ std::vector<float> Helpers::GenerateRandomVector(int size, const Token& token) {
   size_t h = 1125899906842597L;  // prime
 
   if (token.class_id != DefaultClass) {
-    for (int i = 0; i < token.class_id.size(); i++)
+    for (unsigned i = 0; i < token.class_id.size(); i++)
       h = 31 * h + token.class_id[i];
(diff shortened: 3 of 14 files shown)


JeanPaulShapo:
Is it better to rewrite some `for` loops using `for (auto &obj : collection)` version?

omtcyf0:
@JeanPaulShapo generally, нуы. But
-Sometimes index is used (i.e. something like i == whatever)
-AFAIK project might be built with `-std=c++0x`, I'm not sure whether the project aims to target older standards too, so I didn't want to break things.

sashafrey:
@omtcyf0 Thank you for cleaning up the warnings! Let's bring this into master.

My only concern is why it didn't trigger a build in travis (CI system). In bigartm we are running travis for pull request for quite some time, and for the first time I see no CI build. Your other pull request #365 have triggered travis build. Have you done anything differently between this pull request and #365?..  Could you please try to re-create this pull request, and if it still triggers no build then I'll cherry-pick your commits manually.

sashafrey:
@JeanPaulShapo Overall it is OK to use `for (auto &obj : collection)` in BigARTM, as long as it builds for Visual Studio 2012. I believe this is the oldest compiler that we current support. VS 2012 doesn't have complete set of C++11, but it recognize the syntax that you've suggested (and it is already used in BigARTM in many places. However for the purpose of this change (e.g. warnings clenaup) I think `unsigned` is a better approach as it results in a smaller change.

sashafrey:
@omtcyf0 I see lots of warnings during google protobuf compilation - do you know how to disable them? Ideally we should completely disable warnings for 3rdparty projects somewhere in `CMakeLists.txt`. Once it will be less noise I can also go through the warnings produced by Visual Studio and remove as many of those as possible.

omtcyf0:
@sashafrey 

> My only concern is why it didn't trigger a build in travis (CI system). In bigartm we are running travis for pull request for quite some time, and for the first time I see no CI build. Your other pull request #365 have triggered travis build. Have you done anything differently between this pull request and #365?.. Could you please try to re-create this pull request, and if it still triggers no build then I'll cherry-pick your commits manually.

Yes, I was wondering about CI not being triggered for this pull request and being triggered for the #365 . AFAIK I didn't do anything differently. Does recreating means closing this one and opening the same one under number 366?

> I see lots of warnings during google protobuf compilation - do you know how to disable them? Ideally we should completely disable warnings for 3rdparty projects somewhere in CMakeLists.txt. Once it will be less noise I can also go through the warnings produced by Visual Studio and remove as many of those as possible.

Both protobuf and gtest suffer from numerous compiler warnings, silencing/preventing them was on my mind too. I suggest we do `set_property(DIRECTORY 3rdparty PROPERTY COMPILE_OPTIONS -w)` in the upper-level `CMakeLists.txt`. Though this needs some changes in the upper part of CMake config, i.e. compile options should not be set before that as `set_property` does not override them. To sum up, it'll require a bit of hacking, but can be done.

sashafrey:
@omtcyf0 You may try creating another branch (same commits - just `git checkout -b <new_branch_name>`), and open a new pull request from that branch.

sashafrey:
@omtcyf0 
> `set_property(DIRECTORY 3rdparty PROPERTY COMPILE_OPTIONS -w)`

Thanks, I'll give it a try!

sashafrey:
Pull request from #366 was green, so merging this change in.

Issue #369, opened 2015-12-07
Memory issues when training multiple models


sashafrey:
Repeated calls to the following function leads to some kind of memory leak.
```
def trainModel(...):
    model = artm.ARTM(...)
    model.load_dictionary(...)
    model.initialize(...)
                                ............... 
    model.fit_offline(....)
    return model
```
And eventually causes out of memory:
``` 
couldn't lock 16384 bytes of memory (secret_session): Cannot allocate memory
Vector smash protection is enabled.
[3305:3305:1204/211154:ERROR:sync_control_vsync_provider.cc(62)]
glXGetSyncValuesOML should not return TRUE with a media stream counter of 0.
```
Trying ``model._lib.ArtmDisposeMasterComponent(model.master.master_id)`` did not help to resolve this issue.

sashafrey:
This is a duplicate of #382 . Submitter confirmed that `model._lib.ArtmDisposeMasterComponent(model.master.master_id)` actually did resolve the issue.

PR #374, opened 2015-12-09
#369 Memory issues when training multiple models

sashafrey:
Destroy master component in ``__exit__`` to avoid memory leak. Users are supposed to do
```
with artm.ARTM(...) as model:
   # use model
```


sashafrey:
@MelLain @romovpa Is this a good pattern?

Issue #70, opened 2014-12-03
Fix Make install for Linux


bigartm:
Currently `CMakeLists.txt` files are not configured to produce meanunful `make install` scripts for Linux. This should be fixed. If I understand correctly, `make install` should put all relevant BigARTM binaries and shared libraries into `/usr/local/lib` and `/usr/local/bin` folders. However I'm not an expert on Linux. If you have any suggestions on the behavior of `make install` - please share your thoughts...

sashafrey:
@JeanPaulShapo Do you have any thoughts on `make install` for BigARTM? Could you try it out, and suggest a behavior that Linux users would normally expect? Is it important to fix, or this is something people normally shouldn't care about?

JeanPaulShapo:
@sashafrey Ok, I'll try it out. Altough I'm not experienced with normal behaviour of `make install`
(I usually use package managers), seems it's not really hard to figure it out. And, of course, this command is important. For example, we can find it while installing Python interpreter from sources.

sashafrey:
Fixed by #378

PR #378, opened 2015-12-12
Enable installing python interface on Unix systems via 'make install'

JeanPaulShapo:
This pull request solves #70. In particular, now it implements all steps that 'make install' should produce now. These changes work well on my laptop with Fedora 21 and Amazon instance with Ubuntu 14.04 LTS. Question and comments are welcome. Also, feel free to comment if it does't work
on your Linux machine.


sashafrey:
Great job, thanks! My linux machine is also an Ubuntu 14.04 on Amazon, so I'm sure it will work too :)
This pull request is only concerned about python API and its python dependencies, correct? I think we also need to install artm.so and make sure that python API is able to find it. Then python story would be complete, and user can simply delete bigartm build folder after `make install` and python API will stll work.

It would be also great to install bigartm (a CLI executable).

JeanPaulShapo:
@sashafrey Ok, I see it's better to describe what steps 'make install' performs according to this pull request:
0. During 'make' step, 'python setup.py build' both for bigartm package and protobuf package are executed (using auxiliary targets python_build and python_protobuf_build), also protoc executable is copied to 3rdparty/protobuf/src/
1. First of all, it installs protobuf 2.5.1-pre via 'python setup.py install'
2. Next, it installs libartm.so and bigartm into appropriate directories (${PREF}/lib and ${PREF}/bin, where ${PREF} is root directory for install)
3. In the end, it installs bigartm-0.7.3 (current version of bigartm) via 'python setup.py install'.

JeanPaulShapo:
@sashafrey There is one problem for 'bigartm' python package: unfortunately, I didn't manage to install 'numpy' as dependency (there are some compilation errors only when I install 'numpy' as dependency, not as standalone package). That's why I think it's better to mention somewhere about this problem and force user to install numpy beforehand, either via system package manager or using `pip`.

sashafrey:
For `numpy` dependency - I think it is fully OK to ask user to install it. This is a very common task, and I expect most users already have numpy installed.

The steps that you descripes for `make install` is exactly what I've been looking for, however I don't understand how it works... Which file is responsible for (2) and (3) logic. Is it somewhere in CMakeLists.txt or python.py? I'm sure I'm missing something...

JeanPaulShapo:
@sashafrey 
for installing libartm.so - file src/artm/CMakeLists.txt, line 141
for installing bigartm executable - file src/bigartm/CMakeLists.txt, line 30
for installing 'bigartm' python package - file python/CMakeLists.txt, lines 17-24

sashafrey:
@JeanPaulShapo Ok, I see. I didn't know it was already implemented, so I've tried to search only withing the changes from pull request. Thank you!

JeanPaulShapo:
@sashafrey Is it OK to close issue #70?

sashafrey:
@JeanPaulShapo Yes, thank you!

PR #380, opened 2015-12-12
add random seed for Phi matrix

MelLain:


sashafrey:
Looks OK. Later we should remove TopicModel.seed field -- currently this is a hack to pass InitializeModelArgs.seed into ApplyTopicModelOperation

PR #383, opened 2015-12-12
Mel lain dictionaries2

sashafrey:


sashafrey:
My changes address some of the concerns from #358, except 
- In ArtmImportDictionary verify that dictionary file has an extention .dict. If filename does not end with .dict we should throw an error, explaining that "Since BigARTM v0.7.4 function ArtmImportDictionary no longer accepts dictionaries create prior to v0.7.4. Dictionaries should be recreated via ArtmGatherDictionary method, and exported with .dict extension". Yes, this is unreliable, and users might already have a .dict file, or they may choose to rename file. However, this is at least something we can do to avoid a crash. Few releases later we will remove this check and let users specify any name for their dictionaries.  
-  do not delete methods ArtmRequestParseCollection, ArtmReconfigureDictionary and ArtmRequestLoadDictionary methods but throw an exception from them, explaining what users needs to do in order to upgrade their scripts.
- restore all deleted fields in CollectionParserConfig and InitializeModelArgs. In Helper::Validate(...) methods check whether those entries are empty; if they are not empty throw an exception that explains what user needs to do.  Mark the fields as "obsoleted" in messages.proto, so that we remember to delete them in the future.
- Move `TokenInfo` into `master_component.cc`. In fact we can do even better and come some logic from MasterComponent to DictionaryImpl, particularly importing, exporting, filtering and gathering dictionaries should be implemented in DictionaryImpl cass. Then TokenInfo will move to Dictionary.cc.
- Verify dictionary in  ``TEST(CollectionParser, Multiclass)``

And we still need some high-level testing of ARTM.py and bigartm.exe interfaces.

PR #384, opened 2015-12-12
Remove unnecessary commands in Travis CI script

JeanPaulShapo:
1) Now PYTHONPATH variable is unnecessary.
2) python setup.py build for 3rdparty/protobuf/python and copying protoc executable are performed during `make`.


sashafrey:
@JeanPaulShapo It is good to avoid `sudo` because then we are running quicker in Travis. For non-sudo scripts they are using container-based infractructure:
```
This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid and setguid executables.
```

PR #394, opened 2015-12-24
minor docs adjustment

dselivanov:
I just start playing with BigARTM and reading docs. Probably there will be more fixes. 
Is there any contribution guidelines?  Should I create pull requests against master branch? one PR per feature/bug fix? I didn't find any restrictions/guides in [devguide](http://docs.bigartm.org/en/latest/devguide.html)

### DIFF
--- a/docs/tutorials/bigartm_cli.txt
+++ b/docs/tutorials/bigartm_cli.txt
@@ -79,5 +79,5 @@ Then you can use ``bigartm`` as described by ``bigartm --help``:
                                           data in processors
 
     Examples:
-            cpp_client -d docword.kos.txt -v vocab.kos.txt
-            set GLOG_logtostderr=1 & cpp_client -d docword.kos.txt -v vocab.kos.txt
+            bigartm -d docword.kos.txt -v vocab.kos.txt
+            set GLOG_logtostderr=1 & bigartm -d docword.kos.txt -v vocab.kos.txt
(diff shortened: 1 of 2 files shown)


sashafrey:
@dselivanov Thanks for cleaning the docs! Any changes in documentation are very welcome, unfortunately now it is laging far behind... 

>Should I create pull requests against master branch? one PR per feature/bug fix?

Yes, pull requests should be against `master` branch. Ideally it should be one PR per feature or bug, but this is not strict and it would be OK to group small changes (especially in documentation).

Few code guidelines listed here:
http://docs.bigartm.org/en/latest/devguide.html#code-style
(basically, google code style + 120 characters line limit in C++, 100 in python)
In addition we follow `pep8` code style for python, which is enforced by our CI (travis builds). Apart from this I don't think we have a specific contribution guidelines. If you plan to change public stuff (messages.proto, c_interface.h, or any method signature in python code) than I would recommend to create an issue and ask for feedback before coding. Apart from this all internal changes should go easily.

sashafrey:
Btw, have you found examples in [bigartm-book](https://github.com/bigartm/bigartm-book) repot? For instance [here](
 https://github.com/bigartm/bigartm-book/blob/master/junk/cli/BigARTM_CommandLineInferface.ipynb) there is an example of bigartm CLI, and more pages are listed in the [readme](https://github.com/bigartm/bigartm-book).

dselivanov:
thx, got it.

PR #400, opened 2015-12-30
add ability to name batches in correct order

MelLain:


src/artm/core/helpers.cc
```diff
@@ -134,26 +134,28 @@ std::vector<std::string> BatchHelpers::ListAllBatches(const boost::filesystem::p
...
   return uuid;
 }
 
 void BatchHelpers::CompactBatch(const Batch& batch, Batch* compacted_batch) {
   if (batch.has_description()) compacted_batch->set_description(batch.description());
-  if (batch.has_id()) compacted_batch->set_id(batch.id());
+  if (batch.has_id())
+    compacted_batch->set_id(batch.id());
+  else
+    compacted_batch->set_id(boost::lexical_cast<std::string>(boost::uuids::random_generator()()));
```
sashafrey: Do not generate `id` inside `CompactBatch`. It sounds like this function is now fixing / working around problems introduced earlier by other methods. I think doing so will just hide some problems. The input batch (`const Batch& batch` argument) should have an `ID`.
Just keep the original line (`if (batch.has_id()) compacted_batch->set_id(batch.id());`).

src/artm/core/collection_parser.cc
```diff
@@ -139,7 +164,9 @@ void CollectionParser::ParseDocwordBagOfWordsUci(TokenMap* token_map) {
     if (item_id != prev_item_id) {
       prev_item_id = item_id;
       if (batch.item_size() >= config_.num_items_per_batch()) {
-        ::artm::core::BatchHelpers::SaveBatch(batch, config_.target_folder());
+        batch.set_id(boost::lexical_cast<std::string>(boost::uuids::random_generator()()));
+        ::artm::core::BatchHelpers::SaveBatch(batch, config_.target_folder(),
+          config_.name_type() == CollectionParserConfig_NameType_Guid ? batch.id() : batch_name_generator.next_name());
```
sashafrey: Minor suggestion to remove `? :` from all `SaveBatch` calls:
* pass condition `config_.name_type() == CollectionParserConfig_NameType_Guid` to `BatchNameGenerator` constructor, and remember it inside `BatchNameGenerator` class
* add argument `const Batch& batch` to `BatchNameGenerator::next_name()`.

src/artm/core/helpers.cc
MelLain: And who will generate the id?

31.12.2015, 17:37, "sashafrey" <notifications@github.com>:

sashafrey: This must be done in the code that created the batch, passed in as the first argument (`const Batch& batch`).

sashafrey: There are just two places that call `CompactBatch`:
`
bool DataLoader::AddBatch(const AddBatchArgs& args) {
int ArtmSaveBatch(const char* disk_path, int length, const char* batch) {
`
In both cases batch is coming from the user. User must set `batch.id`, and then `CompactBatch` will copy it into `compacted_batch`.

sashafrey:
Merged in pr #405

PR #401, opened 2015-12-30
Master model

sashafrey:


sashafrey:
TBD: restore `reuse_theta`, `disk_cache_folder` and `b_disable_avx_opt` options in BigARTM CLI.

Issue #308, opened 2015-09-25
Allow user to configure DataFrame's column labels in theta matrix


sashafrey:
Currently ``ARTM.transform`` and ``ARTM.fit_transform`` methods use ``ThetaMatrix.item_id`` as a labels of the columns of the resulting DataFrame object. Some users may want to use ``ThetaMatrix.item_title``. This may be particularly useful for VW-format, where user have no control over ``item_id``, while ``item_title`` is set from the first token in each line of the input file.

The code change is really simple. Basically this is about changing from
```
document_ids = [item_id for item_id in theta_info.item_id]
```
to
```
document_ids = [item_id for item_id in theta_info.item_title]
```
in ``python/artm/model.py``, in the code of ``transform`` and ``fit_transform`` methods.

However, the real challenge is to design how the user should select which way to use. Should this be an argument of both methods? Should it be a flag on ARTM class? Or something else?..

sashafrey:
This is done, see ``theta_columns_naming`` option in ``model.py``.

PR #411, opened 2016-01-04
Minor enhancements on phi and theta retrieval in python API

AnyaP:
1. Add model_name to get_phi arguments to be able to extract not only phi, but any word-topic matrix (e.g. nwt counters) in a similar way.
2. Add flag use_doc_titles to fit_transform arguments to name dataframe columns by document titles from Vowpal Wabbit input file.
3. Also I suggest that fit_transform should be renamed to get_theta, because its code is analogous to get_phi. Moreover, it never performs any "fit" (either it returns cached theta, or raises an error), so an analogy with sklearn is misleading.

### DIFF
--- a/python/artm/model.py
+++ b/python/artm/model.py
@@ -77,6 +77,8 @@ class ARTM(object):
       scores(list): list of scores (objects of artm.***Score classes), default=None
       regularizers(list): list with regularizers (objects of
       artm.***Regularizer classes), default=None
+      theta_columns_naming (string): either 'id' or 'title',
+      determines how to name columns (documents) in theta dataframe, default='id'
 
     Important public fields:
       regularizers: contains dict of regularizers, included into model
@@ -96,10 +98,11 @@ class ARTM(object):
 
     # ========== CONSTRUCTOR ==========
     def __init__(self, num_processors=0, topic_names=None, num_topics=10, class_ids=None,
-                 cache_theta=True, scores=None, regularizers=None):
+                 cache_theta=True, scores=None, regularizers=None, theta_columns_naming='id'):
         self._num_processors = 0
         self._num_topics = 10
         self._cache_theta = True
+        self._theta_columns_naming = 'id'
 
         if topic_names is None or not topic_names:
             self._topic_names = []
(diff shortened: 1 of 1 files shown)


python/artm/model.py
```diff
@@ -638,8 +640,11 @@ def get_phi(self, topic_names=None, class_ids=None):
         if not self._initialized:
             raise RuntimeError('Model does not exist yet. Use ARTM.initialize()/ARTM.fit_*()')
 
-        phi_info = self.master.get_phi_info(model=self.model_pwt)
-        nd_array = self.master.get_phi_matrix(model=self.model_pwt,
+        if model_name is None:
+            model_name = self.model_pwt
```
sashafrey:
```
if model_name is None:
  model_name = self.model_pwt
```
I'm not sure if in python this works as you expect it... From what I understand this this code will actually change the default value for `model_name` argument that will affect next calls to `get_phi`. @MelLain , am I correct? How would you suggest to change this?

python/artm/model.py
```diff
@@ -653,7 +658,7 @@ def get_phi(self, topic_names=None, class_ids=None):
 
         return phi_data_frame
 
-    def fit_transform(self, topic_names=None, remove_theta=False):
+    def get_theta(self, topic_names=None, remove_theta=False, use_doc_titles=False):
```
sashafrey: I'd keep old method too for now (ideally tagged with `obsolete`, but I don't know how to do this in python)
```
def fit_transform(self, topic_names=None, remove_theta=False):
  return self.get_theta(topic_names, remove_theta);
```

python/artm/model.py
MelLain: Don't change model_name, it's an argument passed by reference. Use smth like this

final_model_name = model_name if model_name is not None else self.model_pwt
phi_info = self.master.get_phi_info(model=final_model_name)
nd_array = self.master.get_phi_matrix(model=final_model_name)

python/artm/model.py
```diff
@@ -662,6 +667,8 @@ def fit_transform(self, topic_names=None, remove_theta=False):
           default=None (means all topics)
           remove_theta (bool): flag indicates save or remove Theta from model
           after extraction, default=False
+          use_doc_titles (bool): name columns of theta dataframe by document titles,
```
sashafrey: Thank you for fixing #308 , this issue had been opened for way too long :)
Have you consider other ways of customizing the behavior? The issue with adding a Boolean flag to `fit_transform` method is that 
* that this flag needs to be in both methods (`get_theta` and `transform`). 
* most likely users want to configure the behavior once and for all calls. 
I would consider adding a string property on `ARTM` class, named for example `ThetaMatrixColumnsType`. Expected values should be `id` and `title`.

sashafrey:
:+1:

PR #418, opened 2016-01-09
add ability to read and modify dictionaries from python

MelLain:


python/artm/master_component.py
```diff
@@ -95,6 +96,52 @@ def create_dictionary(self, dictionary_data, dictionary_name=None):
 
         self._lib.ArtmCreateDictionary(self.master_id, dictionary_data)
 
+    def save_dictionary_readable(self, filename, dictionary_name, encoding='utf-8'):
```
sashafrey: In my opinion methods `save_dictinoary_readable` and `load_dictionary_readable` belong to `model.py` and should be moved there. `master_component` is just responsible for convenient calls to `wrapper` - nothing else... It makes sense to keep this code here:
```
args = messages.GetDictionaryArgs()  
args.dictionary_name = dictionary_name  
result = self._lib.ArtmRequestDictionary(self.master_id, args)  
dictionary_data = messages.DictionaryData()  
dictionary_data.ParseFromString(result)  
```
I guess this method should be call it `get_dictionary`. And the rest of the functionality should go into model.py.

sashafrey:
@MelLain FYI, github does not send e-mail notifications when you push commits with `-f` (force) option. This is ok, but it would be good to add a dummy comment to the pull request - I believe this will send notififcations.

sashafrey:
@MelLain Oh, I just noticed... next time when you change `master_component.py` please remove `import codecs` - it was introduced by this PR not on purpose. I you still remember.. Otherwise just forget about this, it is not a problem :)

PR #422, opened 2016-01-16
Apply regularizers in artm.transform

sashafrey:


sashafrey:
@MelLain Theta regularizes were not applied in artm.transform... Please rewrite artm python API using MasterModel -- this should help to avoid this sort of issues :) Thanks!

PR #442, opened 2016-02-10
Allow empty Batch.token

sashafrey:
In some scenarios users have a global dictionary at the point where they generate batches. The don't want to replicate the global dictionary across all batches, nor they would like to write code that find a set of all distinct tokens present in the batch and re-index them according to batch dictionary. Ideally the users would like to put token IDs from the global dictionary into `Batch.Item.Field.token_id`, and keep field Batch.token empty. Then the global dictionary can be deployed via ArtmCreateDictionary / ArtmInitializeModel, and BigARTM should treat all batches as if their `Batch.token` field matches with `DictionaryData.token`.

Such batches will have limited usage (for example they must be rejected in ArtmGatherDictionary operation). The fix does some cheating - whenever it sees an empty `Batch.token` it populates it with the entire dictionary. This is not good for performance, but we will improve this later after cleaning up the code in ``processor.cc``.

### DIFF
--- a/src/artm/core/check_messages.h
+++ b/src/artm/core/check_messages.h
@@ -167,7 +167,13 @@ inline std::string DescribeErrors(const ::artm::Batch& message) {
     return ss.str();
   }
 
-  if (message.class_id_size() != message.token_size()) {
+  const bool has_tokens = (message.token_size() > 0);
+  if (!has_tokens && (message.class_id_size() > 0)) {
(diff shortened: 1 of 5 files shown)


sashafrey:
@MelLain What do you think about this change? Do you think this would be a useful scenario to support empty `Batch.token`, and let user point his `Field.item_id`s straight into `DictionaryData.token`?

PR #448, opened 2016-02-13
add log configuring

MelLain:


src/artm/messages.proto
```diff
@@ -846,3 +846,8 @@ message TransformMasterModelArgs {
   optional ThetaMatrixType theta_matrix_type = 3 [default = Dense];
   optional string predict_class_id = 4;
 }
+
+message ConfigureLoggingArgs {
```
sashafrey: message ConfigureLoggingArgs {
  enum Level {
    Info = 0,
    Warning = 1,
    Error = 2,
    Fatal = 3,   
  }
  optional Level minloglevel = 1;
  optional string log_dir = 2;
}

MelLain: I thought about such enum, but
1) the level can be any int, not only 0-3
2) log level has no dependence with error level, 3 means printing all messages, there's nothing fatal in it)

14.02.2016, 01:30, "sashafrey" <notifications@github.com>:

MelLain: But this enum can be used, if levels will be smth like 'Low', 'Medium', 'High' and 'All'

14.02.2016, 01:30, "sashafrey" <notifications@github.com>:

src/artm/c_interface.cc
```diff
@@ -57,19 +57,38 @@ static void set_last_error(const std::string& error) {
   last_error_->assign(error);
 }
 
-static void EnableLogging() {
+static void EnableLogging(artm::ConfigureLoggingArgs* args_ptr = nullptr) {
```
sashafrey: When `EnableLogging` is called through `ArtmConfigureLogging` it is important to report back errors --- not `ARTM_SUCCESS`. I would achieve this by
1. Creating an new overload method `static void EnableLogging(const artm::ConfigureLoggingArgs& args)`; the existing method `static void EnableLogging()` is then kept unchanged.
2. remote `try` and `catch` from the overloaded method
3. in `ArtmConfigureLogging` do
```
try {
  ::artm::ConfigureLoggingArgs args;
  ParseFromArray(configure_logging_args, length, &args);
  EnableLogging(&args);
  return ARTM_SUCCESS;
} CATCH_EXCEPTIONS;
```

src/artm/messages.proto
sashafrey: > the level can be any int, not only 0-3
Ok, I agree -- then enum might not be good idea. Let's keep int. Maybe you may copy a line from glog help into messages.proto just before the `minloglevel` field.
```// The numbers of severity levels INFO, WARNING, ERROR, and FATAL are 0, 1, 2, and 3, respectively.```

PR #452, opened 2016-02-17
Change low-level Python API to use MasterModel

MelLain:


python/artm/master_component.py
```diff
@@ -6,61 +6,173 @@
...
+            master_config.cache_theta = cache_theta
+
+        if pwt_name is not None:
+            master_config.pwt_name = pwt_name
+
+        if nwt_name is not None:
+            master_config.nwt_name = nwt_name
+
+        if pwt_name is not None:
+            master_config.pwt_name = pwt_name
```
sashafrey: `pwt_name` is set twice (here and few lines above)

python/artm/master_component.py
```diff
@@ -631,3 +690,113 @@ def get_info(self):
...
+        if theta_matrix_type is not None:
+            args.theta_matrix_type = theta_matrix_type
+
+        if predict_class_id is not None:
+            args.predict_class_id = predict_class_id
+
+        result = self._lib.ArtmRequestTransformMasterModel(self.master_id, args)
+
+        theta_matrix = messages.ThetaMatrix()
+        theta_matrix.ParseFromString(result)
```
sashafrey: Just curious - why `wrapper.py` doesn't do `theta_matrix = messages.ThetaMatrix()` and `theta_matrix.ParseFromString(result)`? You have to manually do it every time after calling `ArtmRequestXxxx`. Not that it is related to this request, but it would be nice to fix this (unless I'm missing something)

sashafrey:
@MelLain Thanks for this change! I'll do my part with ScoreTracker in C++ over this weekend, so we can finish this pull request asap.

Could you briefly comment on what manual testing have you done so far for this changes in python API? I'm afraid that we have no tests for `ARTM.py`, and it is very likely to get regressions from this kind of refactoring.

sashafrey:
@MelLain Also, could you list all breaking changes in the API, if there is anything else except those two that I've mentioned (changes to `fit_online` signature, and making `topic_names` required in the constructor)

sashafrey:
@MelLain Looks good!
One last question... I think it is common for users to call fit_online repeatedly in a loop, like this:
```
for iter in range(0, 10):
   artm.fit_online(...)
```
Previously this `apply_weight` coefficients were getting smaller and smaller from iteration to iteration (`iter`), while `decay_weight` tend to `1`. Now the behavior will be to use the same vector of coefficients for every iteration, because I think you don't store any state in between fit_online calls (except `_synchronizations_processed`, but it serves a different purpose). 

Could you fix this? Ideally the loop above should behave similarly to CLI `bigartm.exe`. Actually it would be a very nice test to compare the outputs of `bigartm.exe` and our python interface. They should be consistent.

sashafrey:
@MelLain Consider
* removing reset_num_processed_documents
* creating a setter for `_num_online_processed_docs` 
* renaming `_num_online_processed_docs` into `_num_online_processed_batches`
In either case pull request is good :) Just merge it in whenever you think it is ready!

PR #454, opened 2016-02-23
Fix logging and include more fields in ConfigureLoggingArgs

sashafrey:
(the fix was to log an info message in EnableLogging; otherwise it might happen that first call to logging happens from multiple threads simultaneously)


sashafrey:
The issue was that BigARTM.exe logs weird message about logging. It didn't happen on every run, but from time to time the output looked as follows:
```
>bigartm.exe --read-vw-corpus vw.mmro.txt --passes 2
Parsing text collection... OK.
Could not create logging file: Invalid argument
COULD NOT CREATE A LOGGINGFILE 20160218-110741.8648!Gathering dictionary from batches... OK.
Initializing random model from dictionary... Could not create logging file: Invalid argument
COULD NOT CREATE A LOGGINGFILE 20160218-110741.8648!OK.
Number of tokens in the model: 7805
```

PR #456, opened 2016-02-26
Cleanup caching

sashafrey:
- no model_name for theta matrix cache
- CleanThetaCache and CleanScoreCache apis

### DIFF
--- a/python/artm/master_component.py
+++ b/python/artm/master_component.py
@@ -307,10 +307,22 @@ class MasterComponent(object):
         init_args.dictionary_name = dictionary_name
         self._lib.ArtmInitializeModel(self.master_id, init_args)
 
+    def clear_theta_cache(self):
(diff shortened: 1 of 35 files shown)


sashafrey:
@MelLain Please review changes in Python API. There are several parts here.

* First, I replaced `GetThetaMatrixArgs.clean_cache` and `ProcessBatchesArgs.reset_scores` with APIs `ArtmCleanThetaCache` and `ArtmCleanScoreCache`. 
* Second, scores and theta matrix will no longer belong to any `model_name` (=> `GetThetaMatrixArgs.model_name` and `GetScoreValueArgs.model_name` are now removed). It means that from user perspective there will be only one theta matrix within each `MasterComponent`, and only one score value for each score. (very soon I'll add a `ScoreTracker` API to return multiple scores after execution of the Online algorithm; but this is a completely different story -- in either case scores correspond to the same `pwt` matrix, so `model_name` argument is redundant and misleading).

PR #457, opened 2016-02-27
[cleanup] remove loop variables used as const refernces

omtcyf0:
This basic patch introduces few fixes to slightly improve performance and remove redundant copying in for loops where const references can be used.

### DIFF
--- a/src/artm/core/dense_phi_matrix.cc
+++ b/src/artm/core/dense_phi_matrix.cc
@@ -94,7 +94,7 @@ const Token& PhiMatrixFrame::token(int index) const {
 
 google::protobuf::RepeatedPtrField<std::string> PhiMatrixFrame::topic_name() const {
   google::protobuf::RepeatedPtrField<std::string> topic_name;
-  for (auto elem : topic_name_) {
+  for (const auto& elem : topic_name_) {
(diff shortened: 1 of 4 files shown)


sashafrey:
Thanks!
I've merged this into `bigartm/master` branch. Let me know if it makes sense to cherry-pick this for `bigartm/stable`. The difference between branches is as follows:
- `bigartm/stable` will be used to produce `v0.7.x` branches (for example `v0.7.5`, `v0.7.6`, and so on). It will only contain urgent fixes, requested by BigARTM users. All protobuf messages and APIs will be backwards-compatible (at least, that's the goal...)
- `bigartm/master` will be used to produce `v0.8.0` branch. This will happen within few month from now. At the same point `bigartm/stable` will be pulled forward to `bigartm/master`, and from that point onwards `v0.8.x` will be based on `bigartm/stable`.
- This process continues for `v0.9.0`, `v0.10.0`, and so on, until we are ready to release `BigARTM v1.0.0`. At that point we will revisit branching strategy.

In general, all major development will happen in `bigartm/master`. `bigartm/stable` will have no breaking changes, except for the point where we bring `bigartm/stable` forward to `bigartm/master` (at the same time with introducing new major release version).

PR #458, opened 2016-02-27
fixed Matrix constructor-related issue

omtcyf0:
I'm not sure why that gets compiled with gcc and whatever other compilers are used, but Clang surely won't compile this and this actually seems wrong for me.

Clang does output the following diagnostics for this error:

```
error: addition of default argument on redeclaration makes this constructor a default constructor
```

This makes sense if you think about it: which constructor is intended to be called when no arguments are provided?

### DIFF
--- a/src/artm/cpp_interface.cc
+++ b/src/artm/cpp_interface.cc
@@ -281,7 +281,7 @@ void MasterModel::DisposeBatch(const std::string& batch_name) {
 Matrix::Matrix() : no_rows_(0), no_columns_(0), data_() {
 }
 
-Matrix::Matrix(int no_rows = 0, int no_columns = 0) : no_rows_(no_rows), no_columns_(no_columns), data_() {
+Matrix::Matrix(int no_rows, int no_columns) : no_rows_(no_rows), no_columns_(no_columns), data_() {
   if (no_rows <= 0 || no_columns <= 0)
     throw ArgumentOutOfRangeException("no_rows and no_columns must be positive");
 


sashafrey:
@omtcyf0 Thank you! Sadly neither microsoft compiler (on my machine) nor gcc (at travis CI) didn't fire alarm for this issue :(

omtcyf0:
@sashafrey well, here's an issue: they shouldn't.

The C++11 Standard **allows** this piece of code to be compiled (refer to the "overload resolution" section). However, adding `Matrix matrix;` somewhere will and should raise `call of overloaded ‘Matrix()’ is ambiguous` in GCC and (I guess) any other reasonable compiler (though honestly I'm not a fan of MSVC because of it continuously violating the Standard both by not implementing its features and adding some totally weird rules which allow insane constructions not only to be compiled, but also to execute without termination).

Clang folks decided not to allow such pieces of code to be compiled even though they were aware of it being allowed by the Standard. There was a discussion in the mailing lists and there is a [C++ Defect Report Page](http://clang.llvm.org/cxx_dr_status.html) where similar issues are going to.

The basic question "why the hell would the Standard ever allow such thing?" is what I'm having in my mind, but sadly I'm not the one working on the Standard :)

From what I know even C++14 Standard does allow this thingy.

The C++ Standard is a strange thing sometimes. For example, the [depr.c.headers] section of the C++11 Standard states that there are 25 headers deprecated in C++11 while in the table below there are 26... C++14 has this exact issue fixed though.

You never stop wondering about C++ :)

PR #474, opened 2016-03-10
Multiple weighted data paths in batch_vectorizer

AnyaP:
Batch_vectorizer now can get lists for data_path and target_folder and weight each group of batches with a weight given by data_weight list. Then these weights go to master.fit_offline and master.fit_online as batch_weights. As a result, a user can easily weight two VW-files or two batch-folders and fit a model on this weighted data.


sashafrey:
@AnyaP Looks ok :) Should I merge, or you would like to also get @MelLain  feedback?

PR #486, opened 2016-03-16
doc fix to remove properties

MelLain:


python/artm/batches_utils.py
```diff
@@ -112,20 +112,35 @@ def __init__(self, batches=None, collection_name=None, data_path='', data_format
...
     def num_batches(self):
+        """
+        :return: the number of batches
+        """
         return len(self._batches_list)
 
     @property
     def data_path(self):
+        """
+        :return: the dick path of batches
```
JeanPaulShapo: @MelLain maybe 'disk path of batches'?

PR #539, opened 2016-04-25
Fix comment language

nkruglikov:
And I wonder what "scres" in that line stands for.


sashafrey:
@nkruglikov Thanks for fixing this!
`scres` is a typo and a mistake. The typo is in the word `scores`. But `scores` is wrong, and the comment should say `regularizers`.

PR #554, opened 2016-05-04
add background tokens ratio score

sashafrey:
Conflicts:
	python/artm/master_component.py
	python/artm/model.py
	python/artm/wrapper/constants.py
	src/artm/messages.proto


sashafrey:
@MelLain Please merge if this looks OK :)

PR #555, opened 2016-05-05
Test utility to verify that c# generated messages

sashafrey:


sashafrey:
@JeanPaulShapo Does this look ok? I will include this in the script that builds BigARTM on our AWS build server that produces release packages -- this will ensure that messages.proto are always compatible with c# proto generator

JeanPaulShapo:
@sashafrey I found no suspected parts of code. However, I'm not familiar with C# :)

PR #556, opened 2016-05-05
Do not generate empty class_id when parsing Vowpal wabbit format (v2)

### LINKED ISSUE #533: Fix regression in collection parser
The issue is due to `if (elem.size() <= 1)` code in collection parser. This code needs to be fixed to handle tokens that have a single character. See
https://github.com/bigartm/bigartm/commit/68873ef498f74200ae89f874d810073e1607029a#commitcomment-17039213 for more info.

sashafrey:
This fixes #533 . Previous fix (https://github.com/bigartm/bigartm/commit/5b43abbe5beb3f3989b21892af7edef98a045235) was wrong as it ignored tokens with length `1`.

### DIFF
--- a/src/artm/core/collection_parser.cc
+++ b/src/artm/core/collection_parser.cc
@@ -372,10 +372,12 @@ void CollectionParser::ParseVowpalWabbit() {
     ClassId class_id = DefaultClass;
     for (unsigned elem_index = 1; elem_index < strs.size(); ++elem_index) {
       std::string elem = strs[elem_index];
-      if (elem.size() <= 1)
+      if (elem.size() == 0)
         continue;
       if (elem[0] == '|') {
         class_id = elem.substr(1);
+        if (class_id.empty())
+          class_id = DefaultClass;
         continue;
       }
 


sashafrey:
@JeanPaulShapo What do you think about this fix?
I think in a long term we should consider taking WV parser from John Langford -- https://github.com/JohnLangford/vowpal_wabbit/blob/master/vowpalwabbit/parse_example.cc
So for now just a simple fix to treat ` | ` (e.g. `space` - `bar` - `space` sequence) as an indication that the subsequent tokens are of `DefaultClass`.

JeanPaulShapo:
@sashafrey I think we can make it better.
What if you use boost::split function with additional argument `token_compress_on` to remove empty tokens while splitting the string? In such case the first condition will be unnecessary.

JeanPaulShapo:
@sashafrey Hmm, it seems that with such modification there can be empty items, but only first or last. Sorry, I misread the definition of `boost::split`.

sashafrey:
@JeanPaulShapo I restored the original version -- I think it is good enough for now. It fixes the specific problem that someone reported recently --- a bar without class should be treated as a default class, not just as empty string.
As I said, it is best if we take VW parser from the original code, because our VW parser has lots of issues. Just looking at their format I see many cases that we don't handle. 
https://github.com/JohnLangford/vowpal_wabbit/wiki/Input-format
To properly support VW parser we need to also write good test cases that provide decent coverage of most corner-case scenarios.

JeanPaulShapo:
@sashafrey Ok, I see.

PR #557, opened 2016-05-05
Switch python module to brand new version 0.7.5

JeanPaulShapo:


sashafrey:
@JeanPaulShapo I'm very close to publish `v0.8.0` and bring `stable` branch forward. Since your fix is for Master branch I think it should already refer to `v0.8.0`.

JeanPaulShapo:
@sashafrey We need to provide the way to automatically get the version of module.

Issue #527, opened 2016-04-07
TopicSelectionThetaRegularizer issues in master branch


sashafrey:
I'm getting the following error from `TopicSelectionThetaRegularizer` in master branch.
We need to fix it and add a unit-test.
```
AttributeError                            Traceback (most recent call last)
<ipython-input-17-c7b07aa7af3c> in <module>()
----> 1 model_artm.fit_offline(batch_vectorizer=batch_vectorizer, num_collection_passes=10)

C:\Users\alexfrey\Documents\GitHub\bigartm\python\artm\model.pyc in fit_offline(self, batch_vectorizer, num_collection_passes, dictionary_filename)
    519         for _ in xrange(num_collection_passes):
    520             # temp code for easy using of TopicSelectionThetaRegularizer from Python
--> 521             _topic_selection_regularizer_func(self, self._regularizers)
    522 
    523             self._synchronizations_processed += 1

C:\Users\alexfrey\Documents\GitHub\bigartm\python\artm\model.pyc in _topic_selection_regularizer_func(self, regularizers)
     55             for value in [n / (e * self.num_topics) if e > 0.0 else 0.0 for e in n_t]:
     56                 config.topic_value.append(value)
---> 57             self.regularizers[name].config = config
     58 
     59 

C:\Users\alexfrey\Documents\GitHub\bigartm\python\artm\regularizers.pyc in config(self, config)
    132     def config(self, config):
    133         self._config = config
--> 134         self._master.reconfigure_regularizer(self._name, self._type, self._config)
    135 
    136 

C:\Users\alexfrey\Documents\GitHub\bigartm\python\artm\master_component.pyc in reconfigure_regularizer(self, name, config, tau)
    520             if regularizer_config.name == name:
    521                 if config is not None:
--> 522                     master_config.regularizer_config[index].config = config.SerializeToString()
    523                 if tau is not None:
    524                     master_config.regularizer_config[index].tau = tau

AttributeError: 'int' object has no attribute 'SerializeToString'
```

sashafrey:
@MelLain I know you had been working on this some time ago. Is this fixed? Do we have a unit-test in python for TopicSelectionThetaRegularizer?

sashafrey:
Theta selection regularizer seems to work well

PR #558, opened 2016-05-05
Topic selection theta tests

sashafrey:


sashafrey:
Adding the tests to verify #527

PR #562, opened 2016-05-09
change random seed and save n_wt in Python

MelLain:


sashafrey:
The change with ``seed`` is OK.
The n_wt part is much more controversial. First, I consider this a breaking change, because your change will break ``load`` method --- with your change ``v0.8.1`` won't be able to load models saved in ``v0.8.0`` because the filename will be different. Second, I think this is a bad hack. If we need to save ``n_wt`` then this needs to be done in the core. The file format needs to be upgraded, so that we can store ``p_wt`` and ``n_wt`` in the same binary file. This will be simpler after we fix https://github.com/bigartm/bigartm/issues/403 .

So, saving ``n_wt`` need to wait until ``v0.9.0``. For now I suggest to write a simple script in your code that exports and re-imports both ``n_wt`` and ``p_wt``-- by calling model.master.import_model(...).

Issue #550, opened 2016-05-03
Fix inconsistent namings around disk, path, folder, file 


sashafrey:
Fix inconsistent namings around disk, path, folder, file in protobuf messages

PR #563, opened 2016-05-11
try to add new input format

MelLain:


MelLain:
I also would like to change some names of fileds in BatchVectorizer. like '_weights' -> '_weights_list' and '_data_path' -> '_batches_folder'

sashafrey:
@MelLain  It would be great if you include a unit-test, and mention this change in the release notes.
For the rename in BatchVectorizer, we has to wait until v0.9.0 -- as we agreed, no changes in v0.8.X should break user scripts. Please mention your suggestions in this ticket:
https://github.com/bigartm/bigartm/issues/550

sashafrey:
@MelLain We just keep adding more and more parameters into BatchVectorizer constructor, which I think is a bad pattern. A better way is to create a separate class for each data format -- for example, artm.UciBowVectorizer, artm.VowpalWabbitVectorizer, artm.NwdVectorizer, etc. For now let's not change BatchVectorizer, but for the new code I think we should really put it into a new class. This will make it easier to review the change, because right now I can't say if you change is safe or not. Significant changing the code in BatchVectorizer require new unit tests  -- not just for new feature, but also for existing formats which aren't covered at the moment.

I also think that if the user has the data in memory then we shouldn't save it to disk in VW format. Your new vectorizer should create in-memory batches, and fit_offline / fit_online methods transfer the batches into BigARTM via memory (via ArtmImportBatches). Do you think it would be a lot more work to implement this approach?

sashafrey:
@MelLain Please mention this change in the release notes (at least one line that lists all new new arguments in BatchVectorizer constructor). The change is for ``v0.8.1``.

PR #566, opened 2016-05-16
Give regularizer access to phi & test it with new regularizer.

bt2901:
Expanded regularizer_interface to allow access to Phi matrix. Created a regularizer loosely based on an iTopicModel (see http://www.ccs.neu.edu/home/yzsun/papers/icdm09_iTopicModel.pdf for details). 

WIP, but feedback is welcome.


sashafrey:
@bt2901 This change needs few fixes to pass Travis-CI build. First it runs cpplint (googe's code style checks). If they pass Travis-CI will also build and run unit-tests (both C++ and python). 

Current cpplint failures:
```
https://travis-ci.org/bigartm/bigartm/builds/130487072#L554
src/artm/regularizer_interface.cc:19:  Should have a space between // and comment  [whitespace/comments] [4]
src/artm/regularizer_interface.h:89:  "private:" should be preceded by a blank line  [whitespace/blank_line] [3]
```

src/artm/messages.proto
```diff
@@ -163,6 +163,11 @@ message TopicSelectionThetaConfig {
   repeated float alpha_iter = 3;
 }
 
+// Represents a configuration of a iTopic-like modal Theta regularizer
+message iTopicThetaConfig {
```
sashafrey: Do you have potential alternatives for the name of this regularizer?
To follow code style it must start with capital letter -- both here in messages.proto, and also the name of C++ class.
But ``ITopicThetaConfig`` is also bad name -- doesn't look like it tells anything about the purpose of this class. In addtion it collides with C# syntax where ``I`` as a prefix means an interface (e.g. purely abstract class).

src/artm/version.h
```diff
@@ -0,0 +1,9 @@
+// Copyright 2016, Additive Regularization of Topic Models.
```
sashafrey: Please exclude this file from your check-in. It is auto-generated as part of the build system. Feel free to add it into ``.gitignore``.

src/artm/regularizer_interface.cc
```diff
@@ -3,11 +3,23 @@
...
 
 std::shared_ptr< ::artm::core::Dictionary> RegularizerInterface::dictionary(const std::string& dictionary_name) {
   return ::artm::core::ThreadSafeDictionaryCollection::singleton().get(dictionary_name);
 }
 
+std::shared_ptr<const ::artm::core::PhiMatrix> RegularizerInterface::GetPhiMatrix(const std::string& model_name) {
+  return instance_->GetPhiMatrixSafe(model_name);
+}
+
+//std::shared_ptr<Score> ScoreCalculatorInterface::CalculateScore() {
```
sashafrey: Please remove commented out code :0

src/artm/regularizer/itopic_theta.cc
```diff
@@ -0,0 +1,72 @@
...
+
+std::shared_ptr<RegularizeThetaAgent>
+iTopicTheta::CreateRegularizeThetaAgent(const Batch& batch,
+                                                const ProcessBatchesArgs& args, double tau) {
+  iTopicThetaAgent* agent = new iTopicThetaAgent();
+  std::shared_ptr<iTopicThetaAgent> retval(agent);
+
+  const int topic_size = args.topic_name_size();
+  const int item_size = batch.item_size();
+  agent->mybatch = batch;
```
sashafrey: Does this even compile? I see that ``mybatch`` is a ``const Batch&``, e.g. reference type. I though this must be set in constructor initializer list. I'm quite sure some compilers will enforce this limitation.

bt2901: it doesn't :)
fixed, added this to constructor

src/artm/messages.proto
bt2901: I was planning to implement the alternative mode for regularization later (so, it will have 'iTopic mode' and 'netPLSA mode'), and then rename it to something like ModalGraphTheta. 

I agree that name sounds mysterious, which is a bad thing.

python/artm/master_component.py
```diff
@@ -23,6 +23,8 @@ def _regularizer_type(config):
         return constants.RegularizerType_SmoothPtdw
     elif isinstance(config, messages.TopicSelectionThetaConfig):
         return constants.RegularizerType_TopicSelectionTheta
+    elif isinstance(config, messages.iTopicThetaConfig):
+        return constants.RegularizerType_iTopicTheta
```
bt2901: By the way, shouldn't here be something like this?
```
else:
    raise NotImplementedError 
```

sashafrey: @bt2901 Sorry, I missed this question. Sure, `` raise NotImplementedError`` is a good idea. I'm not sure how this normally expressed in python -- so choose any pythonic way, up to you :)

ofrei:
@bt2901 Is there anything that I can help with on this PR? I think we had a long discussion on e-mail and there were some things that you worked on your side. Let me know if you have a chance to follow up on this PR, otherwise I will have to close it.

ofrei:
Closed without merging.

PR #571, opened 2016-05-22
Add LDA model python API. Step 1

MelLain:


MelLain:
This is untested code for discussion. All tests will be added in further commit.

MelLain:
Two minor problems: add parameter theta_matrix_type into transform() and set alpha and beta non-zero in the test

Issue #594, opened 2016-06-14
v 8.0 build failed on Mac OS


MelLain:
Attempt to build version 8.0 of the  library on Mac OS fails with following output:

clang: warning: argument unused during compilation: '-pthread'
ld: library not found for -lcrt0.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [src/bigartm/bigartm] Error 1
make[1]: *** [src/bigartm/CMakeFiles/bigartm.dir/all] Error 2
make: *** [all] Error 2

sashafrey:
Most build issues in ``bigartm`` (CLI) I've seen so far were solved by turning off static linkage (``cmake -DBUILD_STATIC_BIGARTM=OFF ..``). Have you tried this?

MelLain:
It seems, that this fix has helped, thanks!

JeanPaulShapo:
@MelLain Can you reproduce this bug in the clean build directory, doing the same steps?

PR #595, opened 2016-06-14
Prevent using out-of-date cache for builld

JeanPaulShapo:


JeanPaulShapo:
Fixes #594

JeanPaulShapo:
It was discovered that build instructions were OK, because if you have non-empty `build` folder before installation, you can't create it again.

PR #597, opened 2016-06-14
Pip install, basic support

JeanPaulShapo:
Right now we have test package on https://testpypi.python.org/pypi.
You can try to install package using typing `pip install -i https://testpypi.python.org/pypi`.
However, you still must install all dependencies except git (python, cmake, boost etc.)
Right now it doesn't work on Windows.
On my machine with *Fedora 22* installation works, but I'd like to receive comments from users with other Linux distributions and Mac OS'es.
Feel free to write about any problems encountered during testing.

PR #598, opened 2016-06-15
add script for creating cooc files from batches + minor fixes

MelLain:


MelLain:
This file can be used for testing (.txt -> .py)
[test.txt](https://github.com/bigartm/bigartm/files/316050/test.txt)

sashafrey:
@MelLain I trust you that it runs - I'm sure you've tested it on your machine)
As for where to put this script, I think it should go into ``python/artm/utility`` (new folder under ``python/artm``).
Also, I think this small pull request should wait until we merge #597 (pip install support), because it is a big change and it would be harder to fix merges if @JeanPaulShapo also changed ``.gitignore`` and ``api.py``.

JeanPaulShapo:
@sashafrey Yes, I've changed `api.py`.

MelLain:
No problems, if this work will be done soon (2-3 days), because after that we strongly need new release with biterms, script for them and relative regularization.

16.06.2016, 22:18, "sashafrey" <notifications@github.com>:

PR #599, opened 2016-06-16
New biterms regularizer

MelLain:


python/artm/regularizers.py
```diff
@@ -588,3 +589,36 @@ def __init__(self, name=None, tau=1.0, topic_names=None,
                                       config=config,
                                       topic_names=topic_names,
                                       alpha_iter=alpha_iter)
+
+
+class BitermsPhiRegularizer(BaseRegularizerPhi):
```
ofrei: At the top of ``regularizers.py`` file there is a variable called ``__ALL__``. Either we should remove it (if its not used), or add there new regularizer.

sashafrey:
This pull request is fine, but I'm worried that there is no test for new regularizer. Do you have time to add the test in python? It would be really nice to see on Travis that new regularizer doesn't crash. I think each regularizer should have a test similar to this one:
https://github.com/bigartm/bigartm/blob/master/python/tests/artm/test_regularizer_topic_selection.py
Then we at least know that it runs. And as with the case of topic selection regularizer we also know that "something is wrong" --- for example, it gives non-deterministic results. This is quite important, and I won't be able to spot those kind of issues via code review.

python/artm/regularizers.py
MelLain: This is my mistake, I fixed it and added the note about it into the instructions in our wiki.

I'll add test for this regularizer today or tomorrow.

16.06.2016, 21:54, "Oleksandr Frei" <notifications@github.com>:

PR #604, opened 2016-06-20
Remove install_requires and setup_requires from python/setup.py

sashafrey:


python/setup.py
```diff
@@ -21,12 +21,13 @@ def find_protoc_exec():
     found_args, rest_args = parser.parse_known_args(sys.argv)
     sys.argv = rest_args
     result = vars(found_args).get(argument_prefix, None)
+    if os.path.exists(result):
+        return result
```
JeanPaulShapo: @sashafrey One doesn't simply evaluate `os.path.exists(None)`.
Maybe you want to write `if result and os.path.exists(result):`?
Anyway, I don't understand the reason to refactor this part.

sashafrey:
@JeanPaulShapo I fixed this as ``if result is not None and os.path.exists(result)``; would it be more pythonic to skip ``is not None`` part?

The reason for the rewrite was that I need ``PROTOC`` env variable to take priority over the command-line argument. This is because on my windows build machine I would like to manually specify where the ``protoc.exe`` is located.

As for the second part (remove numpy / panda) --- I won't integrate it of course. The problem was that I didn't have numpy installed on my build machine. In this setup the build will fail in ``python_build`` project.  For now I'll just install Anaconda --- its easy and it comes with numpy.

JeanPaulShapo:
@sashafrey If `result` is not supposed to be `numpy.array`, then it's "better" to skip `is not None` :)
Anyway, good job, I didn't think about such scenario!

PR #605, opened 2016-06-21
Protobuf2.6.1

sashafrey:


sashafrey:
Wow.. I didn't expect this to be that easy :) This change upgrades to protobuf 2.6.1. It also works well on my windows machine. It would be nice to test that we can read old batches, but I'm pretty sure that this is the case as the wire format of google protobuf must not change between 2.5.0 and 2.6.1 (there is a very nice article on the format [here](https://developers.google.com/protocol-buffers/docs/encoding)).

@JeanPaulShapo Should I merge this into master?

JeanPaulShapo:
@sashafrey Yes, definitely.

PR #617, opened 2016-07-20
Configure appveyor (continuous integration for windows)

ofrei:
I've started with https://github.com/JohnLangford/vowpal_wabbit/blob/master/appveyor.yml, which installs boost dependency via nugets. But for BigARTM this turned out to be bad approach because our build system assumes that boost is installed from [Prebuilt Windows Binaries](http://sourceforge.net/projects/boost/files/boost-binaries/). Luckily, this is already available on appveyor build agents.

Then I switched to https://github.com/dartsim/dart/blob/master/appveyor.yml, which turned out to be a really good example.

I've learned that it is possible to login to build agents via RDP, which is incredibly helpful!
http://www.appveyor.com/docs/how-to/rdp-to-build-worker
I kept those lines commented out, so if you even need to use this, please uncomment few lines at the end of our ``appveyor.yml`` script.

Finally I needed some inspiration to setup numpy. A good example was available here:
http://tjelvarolsson.com/blog/how-to-continuously-test-your-python-code-on-windows-using-appveyor/

Now this seems to work well. Next step is to build windows packages on appveyor build agent, and upload them to AWS S3. But this can be done later, so for now this pull request is finished.


ofrei:
@JeanPaulShapo That's it, appveyor :) Should I merge this in?

JeanPaulShapo:
@ofrei It's great! You should definitely merge it!

ofrei:
@JeanPaulShapo Done :)

PR #619, opened 2016-07-22
Fix problem with newline in vocab files

### LINKED ISSUE #519: vocab file in UCI bow require \n in the last line
In python users might choose to export vocab file as follows:
```
"\n".join(words)
```
This won't work because currently the UCI bow parser in BigARTM core require a `\n` also in the last line. So python script must have something like this: ``"\n".join(words) + "\n"``, which is really ugly.
We should fix the parser to properly read the last line without \n.

JeanPaulShapo:
@ofrei
* this PR indents to fix #519;
* new vocab file must be without the last newline;
* is it good idea to have identical code in `src/artm/core/dicitonary.cc` and `src/artm/core/collection_parser.cc`?

### DIFF
--- a/src/artm/core/dictionary.cc
+++ b/src/artm/core/dictionary.cc
@@ -256,7 +256,7 @@ Dictionary::Gather(const GatherDictionaryArgs& args,
       int token_id = 0;
       while (!vocab.eof()) {
         std::getline(vocab, str);
-        if (vocab.eof())
+        if (vocab.eof() && str.empty())
           break;
 
         boost::algorithm::trim(str);
@@ -545,3 +545,4 @@ float Dictionary::CountTopicCoherence(const std::vector<core::Token>& tokens_to_
 
 }  // namespace core
 }  // namespace artm
+// vim: set ts=2 sw=2:
--- a/src/artm_tests/collection_parser_test.cc
+++ b/src/artm_tests/collection_parser_test.cc
@@ -46,36 +46,48 @@ TEST(CollectionParser, UciBagOfWords) {
 
   artm::MasterModelConfig master_config;
   ::artm::MasterModel mc(master_config);
-  artm::GatherDictionaryArgs gather_config;
-  gather_config.set_data_path(target_folder);
-  gather_config.set_vocab_file_path(config.vocab_file_path());
-  gather_config.set_dictionary_target_name("mydictionary");
-  mc.GatherDictionary(gather_config);
 
-  ::artm::GetDictionaryArgs get_dictionary_args;
-  get_dictionary_args.set_dictionary_name("mydictionary");
-  auto dictionary = mc.GetDictionary(get_dictionary_args);
-  ASSERT_EQ(dictionary.token_size(), 3);
+  auto dictionary_checker = [&mc, &target_folder] (
+    const std::string &path_to_vocab,
+    const std::string &dict_name) -> void {
+    // first of all, we gather dictionary into the core
+    artm::GatherDictionaryArgs gather_config;
+    gather_config.set_data_path(target_folder);
+    gather_config.set_vocab_file_path(path_to_vocab);
+    gather_config.set_dictionary_target_name(dict_name);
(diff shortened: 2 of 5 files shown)


src/artm/core/dictionary.cc
```diff
@@ -256,7 +256,7 @@ Dictionary::Gather(const GatherDictionaryArgs& args,
       int token_id = 0;
       while (!vocab.eof()) {
         std::getline(vocab, str);
-        if (vocab.eof())
+        if (vocab.eof() && str.empty())
```
ofrei: Looks good!

src/artm/core/dictionary.cc
```diff
@@ -545,3 +545,4 @@ float Dictionary::CountTopicCoherence(const std::vector<core::Token>& tokens_to_
 
 }  // namespace core
 }  // namespace artm
+// vim: set ts=2 sw=2:
```
ofrei: >// vim: set ts=2 sw=2:
Is this intentional?

src/artm_tests/collection_parser_test.cc
```diff
@@ -46,36 +46,48 @@ TEST(CollectionParser, UciBagOfWords) {
...
+    ASSERT_APPROX_EQ(dict.token_value(0), 5.0 / 18.0);
+    ASSERT_APPROX_EQ(dict.token_value(1), 2.0 / 9.0);
+    ASSERT_APPROX_EQ(dict.token_value(2), 0.5);
+  };
 
-  ASSERT_APPROX_EQ(dictionary.token_value(0), 5.0 / 18.0);
-  ASSERT_APPROX_EQ(dictionary.token_value(1), 2.0 / 9.0);
-  ASSERT_APPROX_EQ(dictionary.token_value(2), 0.5);
+  dictionary_checker(config.vocab_file_path(), "default_dictionary");
+  dictionary_checker("../../../test_data/vocab.parser_test_no_newline.txt", "no_newline_dictionary");
```
ofrei: >../../../test_data/vocab.parser_test_no_newline.txt
``../../..`` in C++ tests is a bad practice, but I see it is used across multiple tests. So feel free to keep this code like this, but if you have time it would be nice to replace replace ``../../../test_data`` with BIGARTM_UNITTEST_DATA env variable. I think that should work, because we anyway have to set this variable for python tests.


It is unfortunate that we start filenames with ``../../..``,

ofrei:
>new vocab file must be without the last newline;
My assumption is that both formats are valid -- both with and without the last newline. Is this the case, or actually "new vocab file **must** be without the last newline;"?

JeanPaulShapo:
@ofrei I crafted it to be **without** the last newline, and in test I checked both files - the old **with** newline and the new **without** newline.
I hope no one will edit this file to add newline.

src/artm/core/dictionary.cc
JeanPaulShapo: Yes, with additions in `.gitignore`.

ofrei:
Ah, I see - if you mean the new test file (``test_data/vocab.parser_test_no_newline.txt``) - sure, it must be without the new line. Thanks for adding a test for this!

ofrei:
>is it good idea to have identical code in `src/artm/core/dicitonary.cc` and `src/artm/core/collection_parser.cc`?
Yes, it's good to keep these two files consistent. Sad that they are copy-pasted like this.

JeanPaulShapo:
I only wonder whether it's good idea to utilize this code in separate function
In that case we only need to fix bugs in one place.

ofrei:
If you could refactor this code to avoid duplication - sure, that would be great.

JeanPaulShapo:
Not in this PR.

ofrei:
Agree)

ofrei:
Sorry, I didn't put [PR:review OK]. You are welcome to fix any of the comments, but in either case this PR is good -- feel free to merge.

JeanPaulShapo:
Whoops, I clicked the wrong button

PR #621, opened 2016-07-24
Try out boost::serialize for serialization in BigARTM

sashafrey:
The plan is to use boost::serialize for dictionaries and models. Serialization of batches may still use google protobuf format.
See https://github.com/bigartm/bigartm/issues/403 for details.


ofrei:
This pull request is re-opened as https://github.com/bigartm/bigartm/pull/629

PR #627, opened 2016-08-10
Support Python3

ofrei:
This pull request contains changes from https://github.com/arxikv/bigartm plus minor changes for windows.


arxikv:
It's alright! Note: in main CMakeLists.txt the default Python command is "python2" instead of "python" because in some nerdy Linux distros (Arch, Manjaro) "python" is a symlink to python3 by default :)

ofrei:
Ok, I see. On windows all versions of python have ``python.exe`` as an executable. If there are multiple versions installed each should go it its own folder. Then user has to either control which python is in his ``PATH``. or use some IDE (like pycharm) that takes care of this.

PR #628, opened 2016-08-26
Add warnings for inconsistent D, W, NNZ in UCI BOW format

### LINKED ISSUE #423: Warn about D, W, NNZ when parsing UCI BOW format
UCI bow format starts with three integer values (`D`, `W` and `NNZ`), which are currently ignored by BigARTM. The values needs to be there according to UCI bow format, but if users don't provide them then it will appears that the library just ignores the first triple `(doc_id, word_id, tf)`.
BigARTM should at least issue a warning when first three values (treated as `D`, `W` and `NNZ`) are inconsistent with the remaining content of the docword file.

JeanPaulShapo:
This PR fixes #423


ofrei:
@JeanPaulShapo Thanks, the fix looks good to me!
Sadly these warnings won't show up in Python. Let me know if you have ideas about connecting python logger to the .info / .warn / .err files produced by glog...

PR #629, opened 2016-08-28
Boost::serialize for data serialization

ofrei:
The plan is to use boost::serialize for dictionaries and models. Serialization of batches may still use google protobuf format.
See https://github.com/bigartm/bigartm/issues/403 for details.


ofrei:
Close without merging. Boost serialization support has pros and cons. Overall there might be net gains of switching to boost serialization, but at the moment we just don't have enough manpower to implement this.

PR #630, opened 2016-08-31
Refactor if-else into declaration-based loops

vmarkovtsev:
...just stumbled upon this and couldn't pass by


ofrei:
@vmarkovtsev Great, thanks for both pull requests!

PR #633, opened 2016-09-03
Implement rand-seed switch to bigartm CLI + small code cleanup

### LINKED ISSUE #319: Implement --rand-seed switch to bigartm CLI
According to BigARTM CLI specification we need to support random seed that allows to control initialization of the topic model.

JeanPaulShapo:
Fixed #319

### DIFF
--- a/src/bigartm/srcmain.cc
+++ b/src/bigartm/srcmain.cc
@@ -241,6 +241,7 @@ struct artm_options {
   std::string topics;
   std::string use_modality;
   std::string predict_class;
+  time_t rand_seed;
 
   // Learning
   int num_collection_passes;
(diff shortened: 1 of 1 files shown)


ofrei:
@JeanPaulShapo I would prefer to rename this switch from ``random-seed`` to just ``seed``.
Currently the default behavior to use time-dependent seed, e.g. different results from run to run.
If user wants to get a deterministic behavior, he should fix a ``seed``. I would use ``random-seed`` in a situation where default behavior is deterministic, and ``--random-seed`` enables time-dependend seed.
Do you agree on this?

ofrei:
@JeanPaulShapo Ah, now I see that the ticket #319 explicitly states in the title that the option should be named ``--random-seed``.  I think this hasn't been planed very well.. Sorry for this.

JeanPaulShapo:
@ofrei You can look into [Russian specification](https://github.com/bigartm/bigartm-book/blob/master/junk/cli/BigARTM_CommandLineInferface.ipynb).

ofrei:
@JeanPaulShapo I see... So we keep ``--random-seed`` then?

JeanPaulShapo:
@ofrei I think yes - we keep `--rand-seed`.

ofrei:
@JeanPaulShapo Agreed! :)

ofrei:
@JeanPaulShapo Forget to mention - it would be nice to mention this change in release notes (https://github.com/bigartm/bigartm/blob/master/docs/release_notes/bigartm_cli.txt).

JeanPaulShapo:
@ofrei Done!

PR #638, opened 2016-09-07
Add CLI parsing progress indication support

### LINKED ISSUE #636: Indicate progress during dataset parsing
Currently, bigartm CLI just prints `Parsing text collection...`. It would be nice if it also printed percentage of the parsing done in case of an interactive terminal stdout. I have 20GB Vowpal Wabbit file and it itakes 30 mins. It would be super cool if it also printed ETA as well.

I will be glad to implement this feature if you agree.

vmarkovtsev:
Used in Vowpal Wabbit collection parsing
Closes #636 

Related changes:
* version.h is generated in the build dir => removed from .gitignore
* MacOSX does not have python2 by default, only python => add APPLE to the conditional
* Two extra whitespaces after "OK." to erase the progress artifacts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,7 +124,5 @@ src/artm/messages.pb.h
 /python/bigartm.egg-info/*
 /python/dist/*
 
-/src/artm/version.h
-
 # configuration file for YouCompleteMe
 .ycm_extra_conf.py
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,9 +36,9 @@ option(BUILD_TESTS "Indicates whether to build artm_tests" ON)
 option(BUILD_BIGARTM_CLI "Indicates whether to build bigartm-CLI executable" ON)
 option(BUILD_INTERNAL_PYTHON_API "Indicates whether to build Python API" ON)
 
-if (MSVC)
+if (MSVC OR APPLE)
 set(PYTHON python CACHE INTERNAL "Python command")
-else (MSVC)
+else (MSVC OR APPLE)
 set(PYTHON python2 CACHE INTERNAL "Python command")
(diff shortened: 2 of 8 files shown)


ofrei:
@vmarkovtsev Thanks! I went through the code and everything looks OK. Nice trick with `isatty(fileno(stderr))` --- I haven't seen it before, and looks very useful! Do you know if it will suppress output to stderr when bigartm runs inside python process? (e.g. user works through python API). I'm mostly curious, otherwise I don't think this is an issue.

It would be great if you could also fix these two things:
- on Windows I'm getting garbadge output:
```
>bigartm.exe -c d:\Datasets\vw.mmro.txt --save-batches tmp
Parsing text collection...    1%←[5D   2%←[5D   3%←[5D   4%←[5D   5%←[5D   6%←[5D   7%←[5D   8%←[5D   9%←[5D
D  87%←[5D  88%←[5D  89%←[5D  90%←[5D  91%←[5D  92%←[5D  93%←[5D  94%←[5D  95%←[5D  96%←[5D  97%←[5D  98%←[5D
```
Would it work on your OS if you replace ``\033[<N>D`` with ``\b``? Note that ``\b`` moves only **one** charactere back, and only withing one line (e.g. it doesn't pass through ``\n``).

- enable the same feature for UCI BOW parser as well, method  ``CollectionParser::ParseDocwordBagOfWordsUci``

vmarkovtsev:
I've updated the code. I had to disable this on Windows (as always, it sucks at standards support). 10 finally got the support but it has to be activated. Since I cannot test it myself, leaving as is.

ofrei:
@vmarkovtsev Thanks! I'll try to fix this for Windows.

PR #644, opened 2016-09-09
Use multiple threads in collection parser (for VW format only)

ofrei:


ofrei:
@JeanPaulShapo @vmarkovtsev I've tested this change on 1.2 GB file (VW format), and on my machine (core i5 with 4 physical CPUs) it reduced parsing time from 340 sec down to 117 sec. Looks useful :)
Does the code look OK? 

P.S. I like the std::async. Actually I've tried to use it in BigARTM long  time ago, but I were stuck due to travis. In our travis.yml it was configured in such a way that I couldn't use neither boost::asyn, nor std::async -- e.g. both boost and the compiler were too old. But now we fixed this, I think at the same time as we introduced python3 support. This is really great!

JeanPaulShapo:
@ofrei I'll check this code ASAP

ofrei:
@JeanPaulShapo I've added a small feature to BigARTM CLI that allows users to re-save their batches back into Vowpal Wabbit format. It is also useful if one wants to convert his UCI Bow into VW format.

vmarkovtsev:
@ofrei BTW std::async is not suited for extensive usage with some stdlib implementations which do not use a thread pool [StackOverflow](http://stackoverflow.com/questions/15666443/which-stdasync-implementations-use-thread-pools). So if each of the tasks takes considerable time, it's OK, but if they are queued often, then one should implement his/her own std;;async on top of some thread pool.

ofrei:
@vmarkovtsev Thanks! I naively expected that ``std::async`` works via thread pool. But for this particular case thread pool is not critical, so I'll keep it as it is.

PR #645, opened 2016-09-10
Impove syntax for adding regularizers

### LINKED ISSUE #512: Re-enterant syntax to adding scores and regularizers
Currently users have to use method `add` to add new regularizer or score, and `operator[]` to edit an existing regularizer or score. This is inconvenient because typically people tend to execute the same cell in their ipython notebook multiple times. This leads to the following:

image

An attempt to assign into the result of operator[] also fails:

image

(and this has another issue - user has to type the name of the score or regularizer twice).

@romovpa @MelLain What is a pythonic way of solving this? Should we just allow second `add` (and overwrite whatever exists)?

Thanks to @nadiinchi for pointing out to this issue!

JeanPaulShapo:
Fixed #512

### DIFF
--- a/python/artm/regularizers.py
+++ b/python/artm/regularizers.py
@@ -1,5 +1,6 @@
 import uuid
 import random
+import warnings
 
 from . import wrapper
 from .wrapper import messages_pb2 as messages
@@ -70,11 +71,17 @@ class Regularizers(object):
         self._data = {}
         self._master = master
 
-    def add(self, regularizer):
-        if regularizer.name not in self._data:
-            self._master.create_regularizer(regularizer.name, regularizer.config, regularizer.tau, regularizer.gamma)
-            regularizer._master = self._master
-            self._data[regularizer.name] = regularizer
+    def add(self, regularizer, overwrite=False):
+        name = regularizer.name
+        if name in self._data and not overwrite:
+            raise AttributeError("Unable to replace existing regularizer.\
+If you really want to do it use overwrite=True argument")
+        # next statement represents ternary operator
+        register_func = (self._master.create_regularizer if name not in self._data else
+                         self._master.reconfigure_regularizer)
+        register_func(name, regularizer.config, regularizer.tau, regularizer.gamma)
+        regularizer._master = self._master
+        self._data[name] = regularizer
 
     def __getitem__(self, name):
         if name in self._data:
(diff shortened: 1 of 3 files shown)


JeanPaulShapo:
Feel free to ask questions.

python/artm/regularizers.py
```diff
@@ -66,29 +67,63 @@ def _update_config(self, obj, first=False):
...
-            self._data[regularizer.name] = regularizer
+        name = regularizer.name
+        if name in self._data and not self._force_add:
+            raise AttributeError("Unable to replace existing regularizer with force_add option turned off.\
+If you really want to replace existing regularizer, set force_add option to True")
+        # next statement represents ternary operator
+        # in C/C++:  `res = cond ? x : y`
+        # in Python: `res = cond and x or y` or
+        #            `res = cond and x or not cond and y` (when x may be equivalent to False)
+        register_func = (name not in self._data and self._master.create_regularizer or
```
ofrei: Why not `x if cond else y`? I though this is the standard way to represent trenary operator in python:
http://stackoverflow.com/questions/394809/does-python-have-a-ternary-conditional-operator

python/artm/regularizers.py
```diff
@@ -66,29 +67,63 @@ def _update_config(self, obj, first=False):
 
 
 class Regularizers(object):
-    def __init__(self, master):
+    def __init__(self, master, force_add=False):
```
ofrei: Instead of `force_add` I would rather create two methods: `add` and `add_or_replace` (first one is strict, second allows to add even it regularizer exists).

python/artm/regularizers.py
JeanPaulShapo: Ouch, this construction was out of my mind last night.
Yes, it's more readable and better.
Sorry for that, I'll fix it now.

python/artm/regularizers.py
JeanPaulShapo: In my opinion, `force_add` intended to be global options that affects all assignment operators, including `__setitem__` (e.g. `model.regularizers['somename'] = SomeReg()`). I don't know where to specify `replace=True` or smth like that when using operator[]-like style.

ofrei: Ah, I see --- I haven't thought about operator[].
But maybe we can agree that `operator[]` always behaves as if `force_add=True`, while `add` always behaves as `force_add=False`? I think this makes sense (add typically means "add something new", so it should raise an error if it already exists; while using operator[] involves assignment operator -- which naturally overwrites previous value).

JeanPaulShapo: It sounds reasonably.
BTW, is it good idea add key `overwite=True` instead of implementing `add_or_replace` method?
I.e. what if we have the following convention:
* `model.regularizers['name'] = SomeReg()` and `model.regularizers.add(..., overwrite=True)` may overwrite previous value.
* `model.regularizers.add(...)` may not ovewrite prevous value, instead it raises exceptions with some useful messages like "... if you want to do it please use overwrite=True".

ofrei: I don't have strong opinion on this, but probably yes... At least `overwrite=True` is better than a seperate method `add_or_replace`. And then this convention would be similar to how many methods in `pandas.DataFrame` deal with ``inplace`` option (for example, [pandas.DataFrame.sort](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.sort.html)).

JeanPaulShapo: Ok, I'll change the behaviour.

ofrei:
@JeanPaulShapo :+1:

PR #647, opened 2016-09-13
sixes in cooc_parser and BatchVectorizer

MelLain:


python/artm/batches_utils.py
```diff
@@ -205,8 +202,8 @@ def __reset_batch():
             for key in global_vocab.keys():
                 global_vocab[key][2] = False  # all tokens haven't appeared in this item yet
 
-            for token_id, value in enumerate(column):
-                value = value.item()
+            col = column if isinstance(column, type(np.zeros([0]))) else column.tolist()[0]
+            for token_id, value in enumerate(col):
```
ofrei: It is hard for me to review the change in ``_parse_nwd`` -- I've never used this format myself. Could you point me to an example of how this function is used? Is it covered by unit tests?

MelLain:
Yes, see test for batch_vectorizer. The main issue is that I use numpy.ndarray in the test and everything was OK. But in my work I've used the matrix, produced by sklearn.CountVectorizer, that has another type. This two formats are the most popular ones, so I added this fix to cover both of them.

14.09.2016, 22:38, "Oleksandr Frei" <notifications@github.com>:

ofrei:
@MelLain Got it, thanks!

ofrei:
(I think we may stop using [PR: review OK] and [PR: ready for review] labels --- GitHub seems to have a built-in support for this :)

ofrei:
@MelLain Could you please try ``squash and merge" option?

MelLain:
Done

ofrei:
@MelLain Thanks. What do you think about squash-merge? I think it is nice to enable this because 
* the history looks cleaner without merge commits
* if we would like to keep two commits in the history then these two are worse two pull requests
* people tend to "push -f" to created a single commit -- with squash this is no longer needed; you can create pull requests with many dirty commits, but they won't pollute the history once you merge;

MelLain:
The last option is especially amazing. I think, that's OK!

14.09.2016, 22:58, "Oleksandr Frei" <notifications@github.com>:

ofrei:
@MelLain cool, I've disabled the `merge` button and kept only the *squash* option. Github is amazing indeed :)

PR #653, opened 2016-10-10
Misc fixes in BigARTM CLI (mostly around dictionaries)

ofrei:


ofrei:
Merge even though Travis is failing.
The reason for travis failure is pandas dependency --- something has change recently outside of our project and now "pip install pandas" doesn't work for us. I've changed our .travis.yml to use miniconda, and it fixed all issues for python 2.7; unfortunately, not for 3.5. I don't have more time to look at it.

PR #658, opened 2016-10-19
dictionary refactoring

MelLain:


ofrei:
Great! I'll add few minor notes, but everything looks OK. 
My major concern is whether you still can load an old dictionary (`.dict` file) saved in previous versions -- e.g. before your refactoring. Especially for large dictionaries (>2 GB) that are split into multiple messages. Have you tested this scenario? Also, not just test that ``ArtmImportDictionary`` don't throw an error, but that the dictionary "works" same as before --- for example, counting coherence still gives same result as before.
Do you have huge dictionaries to test this scenario?

src/artm/core/master_component.h
```diff
@@ -94,8 +95,8 @@ class MasterComponent : boost::noncopyable {
 
   void CreateOrReconfigureRegularizer(const RegularizerConfig& config);
 
+  void AddDictionary(std::shared_ptr<Dictionary> dictionary);
```
ofrei: Move `AddDictionary` to `private:` section? It seems to be implementation detail of master component.

src/artm/core/dictionary_operations.h
```diff
@@ -0,0 +1,35 @@
...
+
+namespace artm {
+namespace core {
+
+// DictionaryOperations contains helper methods to operate on Dictionary class.
+class DictionaryOperations {
+ public:
+  static std::shared_ptr<Dictionary> Create(const DictionaryData& data);
+
+  static void Export(const ExportDictionaryArgs& args, ThreadSafeDictionaryCollection* dictionaries);
```
ofrei: A cleaner signature for ``Export`` would be 
```
static void Export(const ExportDictionaryArgs& args, const Dictionary& dict);
```

src/artm/messages.proto
```diff
@@ -460,6 +460,8 @@ message DictionaryData {
   repeated int32 cooc_second_index = 8;
   repeated float cooc_value = 9;
   optional int64 num_items_in_collection = 10;  // number of documents in the collection
+  repeated float cooc_tf = 11; // these two arrays should be either empty or with size
+  repeated float cooc_df = 12; // equal to cooc_value and consistent with cooc indices
```
ofrei: >these two arrays should be either empty or with size
>equal to cooc_value and consistent with cooc indices

Please verify this condition in ``check_messages.h``. It is full of templates --- do you know how it works and how to use it? It should be very straighforward...

ofrei:
@MelLain please rebase this PR to bigartm/master -- that should also fix CI builds

PR #675, opened 2016-11-09
Update NumPy to use fresh enough version

JeanPaulShapo:


ofrei:
@JeanPaulShapo Super, thanks!

PR #680, opened 2016-11-14
Fix some cmake bugs and warnings for linux

dimack:
1. set bin and lib dirs not only for windows
2. remove -Werror flag on linux in debug mode, cause of many existing warnings, and therefore, anable to build
3. fix cmake warning


ofrei:
@dimack Thanks! I've tested that `make install` works well (that was my only concern).

PR #683, opened 2016-11-17
removed 'append' to 'xrange' when 'num_batches % update_every != 0'

FedyuninV:


ofrei:
@FedyuninV I'm sorry for slow reply, I have missed your pull request.
Great that you've spotted the problem and fixed it. Thanks!

PR #690, opened 2016-12-04
Improve SSTheta reg to use custom regularization

MelLain:
I changed some tests values, I have no ideas why their checking values have changed, let's see the results of CI.

Besides of adding new functionality of regularizer I made some minor refactoring in python api.

### DIFF
--- a/src/artm/core/processor.cc
+++ b/src/artm/core/processor.cc
@@ -57,14 +57,16 @@ class RegularizeThetaAgentCollection : public RegularizeThetaAgent {
 
   bool empty() const { return agents_.empty(); }
 
-  virtual void Apply(int item_index, int inner_iter, int topics_size, const float* n_td, float* r_td) const {
+  virtual void Apply(const std::string& item_title, int inner_iter, int topics_size,
+                     const float* n_td, float* r_td) const {
     for (auto& agent : agents_)
-      agent->Apply(item_index, inner_iter, topics_size, n_td, r_td);
+      agent->Apply(item_title, inner_iter, topics_size, n_td, r_td);
   }
 
-  virtual void Apply(int inner_iter, const LocalThetaMatrix<float>& n_td, LocalThetaMatrix<float>* r_td) const {
+  virtual void Apply(int inner_iter, const LocalThetaMatrix<float>& n_td,
+                     LocalThetaMatrix<float>* r_td, const Batch& batch) const {
     for (auto& agent : agents_)
-      agent->Apply(inner_iter, n_td, r_td);
+      agent->Apply(inner_iter, n_td, r_td, batch);
   }
 };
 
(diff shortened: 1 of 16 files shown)


src/artm/core/processor.cc
```diff
@@ -57,14 +57,16 @@ class RegularizeThetaAgentCollection : public RegularizeThetaAgent {
...
-      agent->Apply(item_index, inner_iter, topics_size, n_td, r_td);
+      agent->Apply(item_title, inner_iter, topics_size, n_td, r_td);
   }
 
-  virtual void Apply(int inner_iter, const LocalThetaMatrix<float>& n_td, LocalThetaMatrix<float>* r_td) const {
+  virtual void Apply(int inner_iter, const LocalThetaMatrix<float>& n_td,
+                     LocalThetaMatrix<float>* r_td, const Batch& batch) const {
     for (auto& agent : agents_)
-      agent->Apply(inner_iter, n_td, r_td);
+      agent->Apply(inner_iter, n_td, r_td, batch);
```
MelLain: instead of passing batch into Apply we can store reference to it in the constructor of the agent, but I don't see any real sence to do it.

ofrei: ThetaAgents exists for performance reasons -- processor.cc calls Regularizer.CreateAgent for each new batch. At this moment regularizer can do any "slow" processing (once per batch), and then agents can be applied quickly in the innermost loop (e.g. when processing given document and inner iteration). 
I reverted changes here:
https://github.com/ofrei/bigartm/commit/b2a7ec72ead61a6168bb840275535d7c7d9ed7e6
And while doing this I found a big issue. You've changed a signature of RegularizerInterface::Apply, but didn't change some implementation --- `hierarchy_sparsing_theta`, `smooth_ptdw`, `topic_segmentation_ptdw`. That's not good :) Could this be a reason for python tests to fail?

MelLain: Ok, I'll change it, but this is not the reason of tests failing.

ofrei: @MelLain I agree -- none of the tests use regularizers that I've mentioned.

MelLain: Maybe it is some OS based problem. I change my OS to Win 10 and tha was the first time I ran tests on it. The test gave different values on my PC and in travis.

ofrei: I don't know.. In my pr #691 all tests pass without changes.

MelLain: And mine also here. The initially problems are on my PC:)

ofrei: I minimize changes down to 6 files in pr #691 Should I merge it (and squash commits into one) if it CI is green?

MelLain:
Yes, merge it!

ofrei:
Merged in #691

PR #691, opened 2016-12-04
Mel lain theta reg - ofrei test

ofrei:


ofrei:
@MelLain I'm just testing why python tests are failing... this is a bit strange.

PR #699, opened 2016-12-09
Disable deprecation warnings from tqdm

ofrei:


ofrei:
@MelLain @AnyaP ok, now I indeed need help in testing this :)
On my machine there was a warning due to `tqdm` library (the one that I used to draw progress bars in ipython notebooks). The warnings were somewhat like this:
```
C:\Program Files\Anaconda3\lib\site-packages\tqdm\_tqdm.py:65: DeprecationWarning: sys.getcheckinterval() and sys.setcheckinterval() are deprecated.  Use sys.setswitchinterval() instead.
  sys.setcheckinterval(100)
```
And they happen each time I run `fit_online`, `fit_offline` or `transform`. 

This pull request makes an ugly hack to disable deprecation warnings from tqdm. 

Could you please let me know if the same issue (e.g. warnings) happen on your machine in latest master branch, and if yes --- is it fixed by this hack?

ofrei:
Merged in pr #700

PR #712, opened 2016-12-30
Make params to regularizers/scores accept single string as parameter …

diver-in-sky:
…for class_ids and topic_names

### DIFF
--- a/python/artm/artm_model.py
+++ b/python/artm/artm_model.py
@@ -587,10 +587,10 @@ class ARTM(object):
         :Description: get custom Phi matrix of model. The extraction of the\
                       whole Phi matrix expects ARTM.phi_ call.
 
-        :param topic_names: list with topics to extract, None value means all topics
-        :type topic_names: list of str
-        :param class_ids: list with class ids to extract, None means all class ids
-        :type class_ids: list of str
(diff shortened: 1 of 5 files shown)


ofrei:
@diver-in-sky Thanks for this PR! I agree with this change, it will be nice to pass a single topic_name or class_id. But I would suggest to use `if isinstance(topic_names, str): topic_names = [topic_names]`. This is safer than `isinstance(topic_names, list)` , because topic_names could be not just list but any iterable object, for example a set. This is the reason why `hARTM` test failed in both Appveyor and Travis personal builds.

Could you make this change and re-submit pull requeste against latest master branch? Also there are minor code style checks failing (`python/artm/artm_model.py:660:44: W291 trailing whitespace`). This is checked by  https://github.com/bigartm/bigartm/blob/master/codestyle_checks.sh

diver-in-sky:
@ofrei Thanks for your feedback! I've updated PR with `six.string_types`, looks like it'll cover one additional case if we compare to `if isinstance(topic_names, str):` – `unicode` string in python2.

ofrei:
@diver-in-sky Great, thank you! The changes are merged into master branch.

PR #721, opened 2017-01-19
More detailed and correct structure of docs. Move datasets and format…

MelLain:
…s into a separate subsection


MelLain:
Will be reopened.

PR #765, opened 2017-02-15
add gethering of cooccurrence dictionary

MichaelSolotky:
Added ability to build 2 dictionaries which contain triples <token_id, token_id, cooccurrence>

### DIFF
--- a/src/bigartm/CMakeLists.txt
+++ b/src/bigartm/CMakeLists.txt
@@ -4,6 +4,7 @@ include_directories(${CMAKE_CURRENT_LIST_DIR}/../)
 
 set(SRC_LIST
 	srcmain.cc
+	cooccurrence_dictionary.cc
 	${CMAKE_SOURCE_DIR}/src/artm/cpp_interface.cc
 )
 
--- a/src/bigartm/cooccurrence_dictionary.cc
+++ b/src/bigartm/cooccurrence_dictionary.cc
@@ -1,0 +1,551 @@
+// Copyright 2017, Additive Regularization of Topic Models.
+
+#include "cooccurrence_dictionary.h"
+
+#include <unordered_map>
+#include <string>
+#include <iostream>
+#include <fstream>
+#include <map>
+#include <vector>
+#include <future>
+#include <mutex>
+#include <sstream>
+#include <iomanip>
+#include <memory>
+
+#include "boost/algorithm/string.hpp"
+#include "boost/filesystem.hpp"
+#include "boost/utility.hpp"
+#include "boost/lexical_cast.hpp"
+#include "boost/uuid/uuid_io.hpp"
+#include "boost/uuid/uuid_generators.hpp"
+
+namespace fs = boost::filesystem;
+
+CooccurrenceBatch::CooccurrenceBatch(int batch_num, const char filemode, const std::string& disk_path) {
+  std::stringstream ss;
+  ss << std::setprecision(4) << batch_num << ".bin";
+  std::string filename = ss.str();
+  fs::path full_filename = fs::path(disk_path) / fs::path(filename);
+  if (filemode == 'w') {
+    out_batch_.open(full_filename.string());
+    if (out_batch_.bad()) {
+      std::cerr << "CooccurrenceBatch::CooccurrenceBatch: Failed to create batch\n";
+      throw 1;
+    }
+  } else {
+    in_batch_.open(full_filename.string());
+    if (in_batch_.bad()) {
+      std::cerr << "CooccurrenceBatch::CooccurrenceBatch: Failed to open existing batch\n";
+      throw 1;
+    }
+  }
+}
+
+<<<<<<< HEAD
+void CooccurrenceBatch::FormNewCell(std::map<int, CoocMap>::iterator& map_node) {
+  cell_.first_token_id = map_node->first;
+  cell_.records.resize(0);
+=======
+void CooccurrenceBatch::FormNewCell(std::map<int, std::map<int, cooccurrence_info>>::iterator &map_node) {
+  cell.first_token_id = map_node->first;
+  cell.records.resize(0);
+>>>>>>> d499b4a... Add files via upload
+  for (auto iter = (map_node->second).begin(); iter != (map_node->second).end(); ++iter) {
+    Triple tmp;
+    tmp.cooc_value = iter->second.cooc_value;
+    tmp.doc_quan   = iter->second.doc_quan;
+    tmp.second_token_id = iter->first;
+    cell_.records.push_back(tmp);
+  }
+  cell_.num_of_triples = cell_.records.size();
+}
+
+int CooccurrenceBatch::ReadCellHeader() {
+<<<<<<< HEAD
+  in_batch_.read(reinterpret_cast<char *>(&cell_.first_token_id), sizeof cell_.first_token_id);
+  in_batch_.read(reinterpret_cast<char *>(&cell_.num_of_triples), sizeof cell_.num_of_triples);
+  if (!in_batch_.eof())
+=======
+  in_batch.read(reinterpret_cast<char *>(&cell), 2 * sizeof(int));
+  if (!in_batch.eof())
+>>>>>>> d499b4a... Add files via upload
+    return true;
+  else {
+    cell_.first_token_id = -1;
+    cell_.num_of_triples = 0;
+    return false;
+  }
+}
+
+void CooccurrenceBatch::ReadRecords() {
+<<<<<<< HEAD
+  cell_.records.resize(cell_.num_of_triples);
+  in_batch_.read(reinterpret_cast<char *>(&cell_.records[0]),
+          sizeof(Triple) * cell_.num_of_triples);
+=======
+  cell.records.resize(cell.num_of_triples);
+  in_batch.read(reinterpret_cast<char *>(&cell.records[0]),
+          sizeof(Triple) * cell.num_of_triples);
+>>>>>>> d499b4a... Add files via upload
+}
+
+int CooccurrenceBatch::ReadCell() {
+  if(ReadCellHeader()) {
+    ReadRecords();
+    return true;
+  }
+  return false;
+}
+
+void CooccurrenceBatch::WriteCell() {
+<<<<<<< HEAD
+  out_batch_.write(reinterpret_cast<char *>(&cell_.first_token_id), sizeof cell_.first_token_id);
+  out_batch_.write(reinterpret_cast<char *>(&cell_.num_of_triples), sizeof cell_.num_of_triples);
+  out_batch_.write(reinterpret_cast<char *>(&cell_.records[0]),
+          sizeof(Triple) * cell_.num_of_triples);
+=======
+  out_batch.write(reinterpret_cast<char *>(&cell), 2 * sizeof(int));
+  out_batch.write(reinterpret_cast<char *>(&cell.records[0]),
+          sizeof(Triple) * cell.num_of_triples);
+>>>>>>> d499b4a... Add files via upload
+}
+
+BatchManager::BatchManager() {
+  batch_quan_ = 0;
+  boost::uuids::uuid uuid = boost::uuids::random_generator()();
+  fs::path dir(boost::lexical_cast<std::string>(uuid));
+  if (fs::exists(dir)) {
+    std::cerr << "Folder with name Co-occurrenceBatch already exists in the working directory, please replace it\n";
+    throw 1;
+  }
+  if (!fs::create_directory(dir)) {
+    std::cerr << "Failed to create directory\n";
+    throw 1;
+  }
+  path_to_batches_ = dir.string();
+}
+
+BatchManager::~BatchManager() { fs::remove_all(path_to_batches_); }
+
+<<<<<<< HEAD
+int BatchManager::GetBatchQuan() { return batch_quan_; }
+
+CooccurrenceBatch* BatchManager::CreateNewBatch() {
+  if (batch_quan_ < max_batch_quan_)
+    return new CooccurrenceBatch(batch_quan_++, 'w', path_to_batches_);
+  else {
+=======
+int BatchManager::GetBatchQuan() { return batch_quan; }
+
+CooccurrenceBatch *BatchManager::CreateNewBatch() {
+  if (batch_quan < max_batch_quan) {
+    return new CooccurrenceBatch(batch_quan++, 'w', path_to_batches);
+  } else {
+>>>>>>> d499b4a... Add files via upload
+    std::cerr << "Too many batches, maximal number of batches = "
+              << max_batch_quan_ << std::endl;
+    throw 1;
+  }
+}
+
+<<<<<<< HEAD
+CooccurrenceBatch* BatchManager::OpenExistingBatch(int batch_num) {
+  return new CooccurrenceBatch(batch_num, 'r', path_to_batches_);
+}
+
+void ResultingBuffer::MergeWithExistingCell(const CooccurrenceBatch& batch) {
+=======
+CooccurrenceBatch *BatchManager::OpenExistingBatch(int batch_num) {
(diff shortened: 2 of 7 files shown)


src/bigartm/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ include_directories(${CMAKE_CURRENT_LIST_DIR}/../)
 
 set(SRC_LIST
```
MelLain: Before merging into master this change should be moved grom src/bigartm into src/artm/core (of course after refactoring). You should create test file for it in src/artm_test and test data into test_data directory.

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+#include "boost/filesystem.hpp"
+#include "boost/utility.hpp"
+
+using namespace std;
+using namespace boost;
+
+CooccurrenceBatch::CooccurrenceBatch(int batch_num, const char filemode, const std::string& disk_path) {
+  const int max_name_len = 30;
+  char num_str[max_name_len] = "";
+  sprintf(num_str,"%.3d", batch_num);
```
MelLain: Everywhere use logging:

LOG(ERROR | WARNING | NOTE) << some text without '\n' at the end

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+  std::string filename = string(num_str);
+  filename += string(".bin");
+  boost::filesystem::path full_filename = boost::filesystem::path(disk_path)
+                 / boost::filesystem::path(filename);
+  if (filemode == 'w') {
+    out_batch.open(full_filename.string());
+    if (out_batch.bad()) {
+      std::cerr << "CooccurrenceBatch::CooccurrenceBatch: Failed to create batch\n";
```
MelLain: LOG(ERROR) <<

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+      throw 1;
+    }
+  } else {
+    in_batch.open(full_filename.string());
+    if (in_batch.bad()) {
+      std::cerr << "CooccurrenceBatch::CooccurrenceBatch: Failed to open existing batch\n";
```
MelLain: LOG(ERROR) <<

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+      throw 1;
```
MelLain: Everywhere you'd like to throw exception use BOOST_THROW_EXCEPTION macro and one of pre-defined artm exceptions

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+  path_to_batches = dir.string();
+}
+
+BatchManager::~BatchManager() {
+  boost::filesystem::remove_all(path_to_batches);
+}
+
+inline int BatchManager::GetBatchQuan() { return batch_quan; }
+
+inline CooccurrenceBatch *BatchManager::CreateNewBatch() {
```
MelLain: Replace everywhere:
type \*var -> type\* var
type &var -> type& var

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+  if (!num_of_threads)
+    num_of_threads = 1;
+  std::vector<std::shared_future<void>> tasks;
+  for (int i = 0; i < num_of_threads; ++i)
+    tasks.push_back(std::move(std::async(std::launch::async, func)));
+  for (int i = 0; i < num_of_threads; ++i)
+    tasks[i].get();
+}
+
+inline void CooccurrenceDictionary::ReadAndMergeBatches() {
```
MelLain: There's no need in inline somewhere in your changes.
Inline has two functions
1) ask compiler to put you function code into the place of call without real call for speed-up (it is really obsolete now, because nowadays compilers can make such decisions by themselves)

2) declaring functions in .h file:
 FYI if you define this function in .h file
> int foo() { return 5; }
and then include it into two .cpp files, you'll get error "Multiple definitions of foo()". To avoid it you can write
> **inline** int foo() { return 5; }

But in this review you have no cases of usage of inline, so remove all it's entries.

src/bigartm/cooccurrence_dictionary.h
```diff
@@ -0,0 +1,151 @@
...
+  int doc_quan;
+  int second_token_id;
+};
+
+struct Pair {
+  double cooc_value;
+  int doc_quan;
+};
+
+struct cooccurrence_info {
```
MelLain: all user-defined type names should have upper-case names "TypeName"

src/bigartm/cooccurrence_dictionary.h
```diff
@@ -0,0 +1,151 @@
...
+  double cooc_value;
+  int doc_quan, prev_doc_id;
+};
+
+using namespace std;
```
MelLain: put all using namespace after headers

src/bigartm/cooccurrence_dictionary.h
```diff
@@ -0,0 +1,151 @@
...
+};
+
+struct Pair {
+  double cooc_value;
+  int doc_quan;
+};
+
+struct cooccurrence_info {
+  double cooc_value;
+  int doc_quan, prev_doc_id;
```
MelLain: one declaration of variable on one line

src/bigartm/cooccurrence_dictionary.h
MelLain: never use "using namespace std"

src/bigartm/cooccurrence_dictionary.h
```diff
@@ -0,0 +1,151 @@
...
+  std::ofstream out_batch;
+
+  CooccurrenceBatch(int batch_num, const char filemode, const string& disk_path);
+ public:
+  Cell cell;
+
+  void FormNewCell(std::map<int, std::map<int, cooccurrence_info>>::iterator& map_node);
+
+  int ReadCellHeader();
+
```
MelLain: Remove empty lines between these four methods: they make code of declaration too long without any profit

src/bigartm/cooccurrence_dictionary.h
```diff
@@ -0,0 +1,151 @@
...
+  int ReadCell();
+
+  void WriteCell();
+};
+
+class BatchManager {
+ private:
+  int batch_quan;
+  const int max_batch_quan = 1000;
```
MelLain: firstly declare public methods, than private methods, than public fields (it's better to declare them private with getters), than private fields

src/bigartm/srcmain.cc
```diff
@@ -1480,6 +1488,9 @@ int main(int argc, char * argv[]) {
 
     po::options_description dictionary_options("Dictionary");
     dictionary_options.add_options()
+      ("cooc-min-tf", po::value(&options.cooc_min_tf)->default_value(200), "minimal value of cooccurrences of a pair of tokens that are saved in dictionary of cooccurrences")
```
MelLain: it's not a good defaults. By defult min_tf and min_df should be both 0.

src/bigartm/cooccurrence_dictionary.h
```diff
@@ -0,0 +1,151 @@
...
+  std::string cooc_df_file_path;
+  bool calculate_tf_cooc;
+  bool calculate_df_cooc;
+  BatchManager batch_manager;
+  std::unordered_map<string, int> dictionary;
+
+  void FetchVocab();
+
+  void UploadBatchOnDisk(BatchManager &batch_manager,
+        std::map<int, std::map<int, cooccurrence_info>>& cooc);
```
MelLain: typedef std::map<int, cooccurrence_info> CoocMap

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+  for (auto iter = cooc.begin(); iter != cooc.end(); ++iter) {
+    batch->FormNewCell(iter);
+    batch->WriteCell();
+  }
+  delete batch;
+}
+
+inline cooccurrence_info CooccurrenceDictionary::FormInitialCoocInfo(int doc_id) {
+  cooccurrence_info res;
+  res.doc_quan = res.cooc_value = 1;
```
MelLain: one assignment in one line

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+}
+
+inline void CooccurrenceDictionary::ModifyCoocMapNode(int second_token_id,
+        int doc_id, std::map<int, cooccurrence_info>& map_node) {
+  auto iter = map_node.find(second_token_id);
+  if (iter == map_node.end()) {
+    cooccurrence_info tmp = FormInitialCoocInfo(doc_id);
+    map_node.insert(std::pair<int, cooccurrence_info> (second_token_id, tmp));
+  } else {
+    iter->second.cooc_value++;
```
MelLain: it's minor, but better use here (++iter->second.cooc_value;) instead

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+  // 6. If resulting map isn't empty create a batch and dump all information
+  // on external storage
+  // Repeat 1-6 for all portions (can work in parallel for different portions)
+
+  std::filebuf fb;
+  if (!fb.open(path_to_vw, std::ios::in)) {
+    std::cerr << "Failed to open vocab\n";
+    throw 1;
+  }
+  std::istream VowpalWabbitDoc(&fb);
```
MelLain: variable names should be in lower case: vowpal_wabbit_doc

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+      }
+
+      if (!portion.size())
+        continue;
+
+      // First elem in external map is first_token_id, in internal it's
+      // second_token_id
+      std::map<int, std::map<int, cooccurrence_info>> cooc_map;
+
+      for (int doc_id = 0; doc_id < (int64_t) portion.size(); ++doc_id) {
```
MelLain: replace (typename) with static_cast<typename>

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+      }
+
+      {
+        std::lock_guard<std::mutex> guard(write_lock);
+        if (!cooc_map.empty())
+          UploadBatchOnDisk(batch_manager, cooc_map);
+      }
+    }
+  };
+  int num_of_threads = std::thread::hardware_concurrency();
```
MelLain: newline here

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+  if (!num_of_threads)
```
MelLain: num_of_threads == 0

Everywhere you have zero or nullptr check use == 0 | nullptr instead of logical !

it more obviously

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+  for (int i = 0; i < num_of_threads; ++i)
+    tasks[i].get();
+}
+
+inline void CooccurrenceDictionary::ReadAndMergeBatches() {
+  auto CompareBatches = [](const CooccurrenceBatch *left,
+                           const CooccurrenceBatch *right) {
+    return left->cell.first_token_id > right->cell.first_token_id;
+  };
+  std::vector<CooccurrenceBatch *> batch_queue;
```
MelLain: Maybe 

std::vector<std::unique_ptr<CooccurrenceBatch> > batch_queue;

to avoid delete in 438 line?

src/bigartm/cooccurrence_dictionary.h
```diff
@@ -0,0 +1,151 @@
+// Copyright 2017, Additive Regularization of Topic Models.
+
+//#include <cstdio>
+#include <unordered_map>
+#include <string>
+#include <iostream>
+#include <cstdlib>
```
MelLain: 1) c includes
2) cpp standart includes
3) cpp 3rdparty includes (like boost)
4) your includes

src/bigartm/cooccurrence_dictionary.h
```diff
@@ -0,0 +1,151 @@
...
+class CooccurrenceBatch: private boost::noncopyable {
+  struct Cell {
+    int first_token_id;
+    int num_of_triples;
+    std::vector<Triple> records;
+  };
+
+  friend class BatchManager;
+ private:
+  std::ifstream in_batch;
```
MelLain: all private fields of all classes should end with '_':

std::ifstream in_batch_;
std::ofstream out_batch_;

src/bigartm/cooccurrence_dictionary.h
```diff
@@ -0,0 +1,151 @@
...
+  void PopPreviousContent();
+
+  void AddNewCellInBuffer(const CooccurrenceBatch *batch);
+ public:
+  ResultingBuffer(const double min_tf, const int min_df,
+          const std::string &cooc_tf_file_path,
+          const std::string &cooc_df_file_path, const bool &cooc_tf_flag,
+          const bool &cooc_df_flag);
+
+  ~ResultingBuffer();
```
MelLain: No need to declare empty non-virtual destructor

MichaelSolotky: It's not empty. Here is definition.
https://github.com/MichaelSolotky/bigartm/blob/master/src/bigartm/cooccurrence_dictionary.cc#L194

MelLain: Oh, sorry, I didn't noticed(

src/bigartm/cooccurrence_dictionary.cc
```diff
@@ -0,0 +1,442 @@
...
+
+inline void CooccurrenceBatch::WriteCell() {
+  out_batch.write(reinterpret_cast<char *>(&cell), 2 * sizeof(int));
+  out_batch.write(reinterpret_cast<char *>(&cell.records[0]),
+          sizeof(Triple) * cell.num_of_triples);
+}
+
+BatchManager::BatchManager() {
+  batch_quan = 0;
+  boost::filesystem::path dir("Co-occurrenceBatches");
```
MelLain: Create random name using uuid

src/bigartm/cooccurrence_dictionary.cc
MichaelSolotky: Ok.
Sorry. It's because of codestyle or there are other reasons?

MelLain: http://stackoverflow.com/questions/24853/what-is-the-difference-between-i-and-i

It can give problems if object++ is big (because of coping), although I think, that modern compilers can avoid it. But still ++object considered to be better style.

MichaelSolotky: Yeah, I know about this, but there was just double. I'll better replace it with += 1 (I think it looks better)

MelLain: No, ++i or at least keep as is.

MichaelSolotky: Ok)

src/bigartm/CMakeLists.txt
MichaelSolotky: Sorry. I don't know where I should add my header in order to use class from it in srcmain. I thought it's common.h until I looked into it.

MelLain: No, everyth is OK for now, I write this comment for future.

PR #772, opened 2017-02-25
add simple wrapper for online processing of batches in memory

MelLain:
I don't see any reasonable way of using this oppurtunity besides multiply iteration over given set of batches. So I don't think it's a good way to parse vw file into the Python memory and than pass batches to C++. The good case (as I think) is the following:
1) create batches as usual
2) pass them into the C++ memory using BV
3) iterate over them
4) remove them by deleting BV

It is a discussable thing.

### DIFF
--- a/python/artm/batches_utils.py
+++ b/python/artm/batches_utils.py
@@ -75,11 +75,14 @@ class BatchVectorizer(object):
         :type class_ids: list of str or str
         :param bool process_in_memory: process batches from disk or from RAM \
                                        (only if data_format == Batches), model parameter is required
-        :param artm.ARTM model: ARTM instance that will use this vectorizer, is need only if\
-                                process_in_memory == True
+        :param artm.ARTM model: ARTM instance that will use this vectorizer, is required when\
+                                process_in_memory == True. Will be ignored in other cases
         """
         self._remove_batches = False
         self._process_in_memory = process_in_memory and data_format == 'batches' and model is not None
+        if self._process_in_memory != process_in_memory:
+            raise IOError("Correct configuration: process_memory == True + data_format == 'batches' + model != None")
+
         self._model = model
         if data_format == 'bow_n_wd' or data_format == 'vowpal_wabbit' or data_format == 'bow_uci':
             self._remove_batches = target_folder is None
@@ -123,7 +126,7 @@ class BatchVectorizer(object):
 
         if self._process_in_memory:
             for batch in self._batches_list:
-                model.master.remove_batch(batch)
+                self._model.master.remove_batch(batch)
         self._process_in_memory = False
 
     def __exit__(self, exc_type, exc_value, traceback):
--- a/python/artm/batches_utils.py
+++ b/python/artm/batches_utils.py
@@ -38,7 +38,8 @@ class Batch(object):
 class BatchVectorizer(object):
     def __init__(self, batches=None, collection_name=None, data_path='', data_format='batches',
                  target_folder=None, batch_size=1000, batch_name_type='code', data_weight=1.0,
-                 n_wd=None, vocabulary=None, gather_dictionary=True, class_ids=None):
+                 n_wd=None, vocabulary=None, gather_dictionary=True, class_ids=None,
+                 process_in_memory=False, model=None):
         """
         :param str collection_name: the name of text collection (required if data_format == 'bow_uci')
         :param str data_path: 1) if data_format == 'bow_uci' => folder containing\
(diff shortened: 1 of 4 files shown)


python/tests/artm/test_batches_utils.py
```diff
@@ -81,11 +105,9 @@ def test_func():
...
             assert counter == n_wd_num_tokens + 2
 
-            # The following tests are commented out because they don't pass with Python 3
-            # Please, investigate.
-            # assert tokens == n_wd_tokens_list
-            # assert token_tf == n_wd_token_tf_list
-            # assert token_df == n_wd_token_df_list
+            assert set(tokens) == set(n_wd_tokens_list)
+            assert set(token_tf) == set(n_wd_token_tf_list)
+            assert set(token_df) == set(n_wd_token_df_list)
```
JeanPaulShapo: @MelLain Could you provide some brief results of investigation (i.e why we can't compare lists in Python3)?

JeanPaulShapo:
@MelLain Do you want to say that this wrapper is intended to use in case when you have some set of Python batches in memory?

python/artm/batches_utils.py
```diff
@@ -70,8 +73,14 @@ def __init__(self, batches=None, collection_name=None, data_path='', data_format
                                        and if data_weight is list - automatically set to False
         :param class_ids: list of class_ids or single class_id to parse and include in batches
         :type class_ids: list of str or str
+        :param bool process_in_memory: process batches from disk or from RAM \
+                                       (only if data_format == Batches), model parameter is required
+        :param artm.ARTM model: ARTM instance that will use this vectorizer, is need only if\
+                                process_in_memory == True
         """
```
JeanPaulShapo: May be it'll be better if we change 'is need only if' to 'required when'?

python/artm/batches_utils.py
```diff
@@ -70,8 +73,14 @@ def __init__(self, batches=None, collection_name=None, data_path='', data_format
                                        and if data_weight is list - automatically set to False
         :param class_ids: list of class_ids or single class_id to parse and include in batches
         :type class_ids: list of str or str
+        :param bool process_in_memory: process batches from disk or from RAM \
+                                       (only if data_format == Batches), model parameter is required
```
JeanPaulShapo: I didn't get it: do we have to pass some model when process_in_memory == False?

python/tests/artm/test_batches_utils.py
MelLain: No, I only made any check instead of nothing.

python/artm/batches_utils.py
MelLain: Maybe)

python/artm/batches_utils.py
MelLain: No, only in case of process_in_memory == True

python/artm/batches_utils.py
```diff
@@ -70,8 +73,14 @@ def __init__(self, batches=None, collection_name=None, data_path='', data_format
...
+        :param artm.ARTM model: ARTM instance that will use this vectorizer, is need only if\
+                                process_in_memory == True
         """
         self._remove_batches = False
+        self._process_in_memory = process_in_memory and data_format == 'batches' and model is not None
```
JeanPaulShapo: I think we need to raise some errors if someone passes no model and process_in_memory == True, because later in *_parse_batches* we treat them as list of filenames instead of message.Batch objects.

MelLain: Yes, you are right

python/tests/artm/test_batches_utils.py
JeanPaulShapo: Ok, I understand you. Maybe we leave some FIXME notice to investigate it later?

python/artm/batches_utils.py
JeanPaulShapo: Is it OK to add note about process_in_memory == True?

python/tests/artm/test_batches_utils.py
MelLain: OK

python/artm/batches_utils.py
MelLain: Well, yes maybe. I will also create a page into documentation about this case, I think it'd be enough

JeanPaulShapo: Agreed with you.

python/artm/batches_utils.py
```diff
@@ -112,6 +121,11 @@ def __dispose(self):
             shutil.rmtree(self._target_folder)
         self._remove_batches = False
 
+        if self._process_in_memory:
+            for batch in self._batches_list:
+                model.master.remove_batch(batch)
+        self._process_in_memory = False
```
JeanPaulShapo: *model* -> *self._model*?

MelLain: Oh, yes)

JeanPaulShapo:
@MelLain Ah, I totally forgot about [release notes](http://docs.bigartm.org/en/stable/release_notes/python.html#changes-in-python-api)! Could you create section about next release (perhaps v0.8.4) and add information about this option?

MelLain:
It is a very raw change) I'll add info into the changelog after some discussion and corrction of the interfaces.

JeanPaulShapo:
@MelLain I see you.

PR #779, opened 2017-03-18
Initial support for SIMD-instructions

JeanPaulShapo:
It's not the final version, so feel free to ask questions.


ofrei:
@JeanPaulShapo I'm not sure how to handle this PR. There are nice ideas in this PR, it would be sad if they are lost forever in unmerged pull request. On another hand this is not finished, and I'm not sure if merging had any side-effects. Should we just close , and come back to this PR if you or me ever have time to followup?

PR #790, opened 2017-04-08
Use multiple threads for build process in tutorial.

akashin:
Most modern processors support multithreading, and people usually just copy the installation command. Let's make installation faster and user's experience smoother!

### DIFF
--- a/docs/installation/linux.txt
+++ b/docs/installation/linux.txt
@@ -46,7 +46,7 @@ Scroll further below for more OS-specific insructinos.
     cd bigartm
     mkdir build && cd build
     cmake ..
-    make
+    make -j20
 
     # Step 4. Install BigARTM
     make install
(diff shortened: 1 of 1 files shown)


docs/installation/linux.txt
```diff
@@ -46,7 +46,7 @@ Scroll further below for more OS-specific insructinos.
     cd bigartm
     mkdir build && cd build
     cmake ..
-    make
+    make -j20
```
JeanPaulShapo: @akashin Why not simply -j?
P.S.: I think someone with 4GB of RAM would be very upset :)

ofrei: @akashin I would keep this as is. Here is another project that prefer to keep instructions as short as possible: https://github.com/google/glog . I would say on Linux users are themselves responsible to run make in parallel.

akashin: I realized that another point for plain `make` would be the fact that in case of problems during build the error output would be much more sane.

I was a bit upset by the fact that I needed to stop running build process and rerun it with higher parallelism. I expected that this codeblock will use the most suitable approach for me.

akashin:
The limited RAM argument is good, gcc can easily blow up the memory usage. Let's keep defaults simple and safe.

PR #798, opened 2017-05-11
Improve decorrelator

MelLain:
- add ability to make custom decorrelation
- fix great bug with topic_names update from Python API

### DIFF
--- a/python/artm/regularizers.py
+++ b/python/artm/regularizers.py
@@ -8,7 +8,7 @@ from . import wrapper
 from .wrapper import messages_pb2 as messages
 from .wrapper import constants as const
 
-from six import string_types
+from six import string_types, iteritems
 
 __all__ = [
     'KlFunctionInfo',
@@ -229,7 +229,7 @@ class BaseRegularizerPhi(BaseRegularizer):
 
     @topic_names.setter
     def topic_names(self, topic_names):
-        _reconfigure_field(self, topic_names, 'topic_names')
+        _reconfigure_field(self, topic_names, 'topic_name')
 
 
 class BaseRegularizerTheta(BaseRegularizer):
@@ -269,7 +269,7 @@ class BaseRegularizerTheta(BaseRegularizer):
 
     @topic_names.setter
     def topic_names(self, topic_names):
-        _reconfigure_field(self, topic_names, 'topic_names')
+        _reconfigure_field(self, topic_names, 'topic_name')
 
 
 ###################################################################################################
@@ -429,7 +429,19 @@ class DecorrelatorPhiRegularizer(BaseRegularizerPhi):
     _config_message = messages.DecorrelatorPhiConfig
     _type = const.RegularizerType_DecorrelatorPhi
 
-    def __init__(self, name=None, tau=1.0, gamma=None, class_ids=None, topic_names=None, config=None):
+    def _update_config(self, pairs):
+        self._config.ClearField('first_topic_name')
+        self._config.ClearField('second_topic_name')
+        self._config.ClearField('value')
+
+        for first_topic, topics_and_values in iteritems(pairs):
+            for second_topic, value in iteritems(topics_and_values):
+                self._config.first_topic_name.append(first_topic)
+                self._config.second_topic_name.append(second_topic)
+                self._config.value.append(value)
+
+    def __init__(self, name=None, tau=1.0, gamma=None, class_ids=None,
+                 topic_names=None, topic_pairs=None, config=None):
         """
         :param str name: the identifier of regularizer, will be auto-generated if not specified
         :param float tau: the coefficient of regularization for this regularizer
@@ -440,6 +452,10 @@ class DecorrelatorPhiRegularizer(BaseRegularizerPhi):
         :param topic_names: list of names or single name of topic to regularize,\
                             will regularize all topics if empty or None
         :type topic_names: list of str or single str or None
+        :param topic_pairs: information about pairwise topic decorralation coefficients,\
+                            all topic names from topic_names parameter will be used with\
+                            1.0 coefficietn if None.
+        :type topic_pairs: dict, key - topic name, value - dict with topic names and float values
         :param config: the low-level config of this regularizer
         :type config: protobuf object
         """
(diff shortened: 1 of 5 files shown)


python/artm/regularizers.py
```diff
@@ -269,7 +269,7 @@ def alpha_iter(self, alpha_iter):
 
     @topic_names.setter
     def topic_names(self, topic_names):
-        _reconfigure_field(self, topic_names, 'topic_names')
+        _reconfigure_field(self, topic_names, 'topic_name')
```
ofrei: `topic_names` -> `topic_name` is a bug fix and this never worked before?

python/tests/artm/test_regularizer_decorrelator.py
```diff
@@ -0,0 +1,78 @@
...
+        real_values = [
+            [0.0,  0.91, 0.21, 0.54, 0.35],
+            [0.0,  0.0,  0.55, 0.26, 0.53],
+            [0.0,  0.08, 0.24, 0.20, 0.12],
+        ]
+
+        for elems, values in zip(phi.values.tolist(), real_values):
+            for e, v in zip(elems, values):
+                assert abs(e - v) < tolerance
+    finally:
```
ofrei: Sorry, I have no idea why this fails on Appveyor but not on Travis
```
            for elems, values in zip(phi.values.tolist(), real_values):
                for e, v in zip(elems, values):
>                   assert abs(e - v) < tolerance
E                   assert 0.9010997086763382 < 0.05
E                    +  where 0.9010997086763382 = abs((0.048900291323661804 - 0.95))
artm\test_regularizer_decorrelator.py:45: AssertionError
==================== 1 failed, 21 passed in 133.09 seconds ====================
Command exited with code 1
```

src/artm/regularizer/decorrelator_phi.cc
```diff
@@ -20,11 +21,19 @@ bool DecorrelatorPhi::RegularizePhi(const ::artm::core::PhiMatrix& p_wt,
   const int topic_size = p_wt.topic_size();
   const int token_size = p_wt.token_size();
 
-  std::vector<bool> topics_to_regularize;
-  if (config_.topic_name().size() == 0)
-    topics_to_regularize.assign(topic_size, true);
-  else
-    topics_to_regularize = core::is_member(p_wt.topic_name(), config_.topic_name());
+  std::unordered_map<std::string, int> topics_to_regularize;
```
ofrei: These changes in decorrelator_phi.cc look OK. Most importantly regularizer should work as before if new fields (first_topic_name, second_topic_name, value) are empty --- which to me seems to be the case (plus this should be already covered with unit-tests).

src/artm/regularizer/decorrelator_phi.cc
```diff
@@ -37,18 +46,36 @@ bool DecorrelatorPhi::RegularizePhi(const ::artm::core::PhiMatrix& p_wt,
...
-      for (int topic_id = 0; topic_id < topic_size; ++topic_id) {
-        if (topics_to_regularize[topic_id]) {
-          float weight = p_wt.get(token_id, topic_id);
-          float value = static_cast<float>(- weight * (weights_sum - weight));
-          result->set(token_id, topic_id, value);
+      for (auto& pair : topics_to_regularize) {
+        if (topic_pairs_.size()) {
+          // NOTE: we regularize only topics from topics_to_regularize, and in
+          // this case they will be regularized using their topics from topic_map_,
+          // even if these topics are not presented in topics_to_regularize.
```
ofrei: To me this behavior is conter-intuitive. I would suggest to require an empty topic_name field in the config if user wants to use fisrt/second_topic_name nad value. Empty topic_name will be populated with all topics, and then your code works very straightforward.

src/artm/regularizer/decorrelator_phi.cc
```diff
@@ -63,6 +90,33 @@ google::protobuf::RepeatedPtrField<std::string> DecorrelatorPhi::class_ids_to_re
...
+  config_.clear_first_topic_name();
+  config_.clear_second_topic_name();
+  config_.clear_value();
+
+  topic_pairs_.clear();
+  int topics_len = config.first_topic_name_size();
+  if (topics_len) {
+    if (topics_len != config.second_topic_name_size() ||
+      topics_len != config.value_size()) {
+      BOOST_THROW_EXCEPTION(::artm::core::CorruptedMessageException(
```
ofrei: Simple validation is best put in check_messages.h , especially things like consistent field length in a protobuf message. You may follow the same pattern as here:
```
// inline void FixMessage(::artm::ScoreConfig* message) {
  if (message->type() == ScoreType_TopTokens)
    FixPackedMessage<TopTokensScoreConfig>(message->mutable_config());
```

python/artm/regularizers.py
MelLain: yes

src/artm/regularizer/decorrelator_phi.cc
MelLain: Ok, I'll move it

MelLain: No, this is not a fix, it'a a description of error. Also I don't want to move into descr section, because it is not a message, that can pass directly through c_interface.

Error checking in SmoothSparsePhi is the same as here.

ofrei: No problem, just keep in here. But in general check_messages.h can be used for two purpose: (1) fixing messages, and (2) throwing error if they can't be fixed.

python/artm/regularizers.py
```diff
@@ -452,14 +468,29 @@ def __init__(self, name=None, tau=1.0, gamma=None, class_ids=None, topic_names=N
...
+            self._topic_pairs = topic_pairs
+
     @property
     def dictionary(self):
         raise KeyError('No dictionary parameter')
 
+    @property
+    def topic_pairs(self):
+        return self.topic_pairs
+
```
JeanPaulShapo: Maybe `self._topic_pairs`?

MelLain: Yes!

MelLain: Fixed

src/artm/regularizer/decorrelator_phi.cc
```diff
@@ -19,12 +20,23 @@ bool DecorrelatorPhi::RegularizePhi(const ::artm::core::PhiMatrix& p_wt,
...
-  else
-    topics_to_regularize = core::is_member(p_wt.topic_name(), config_.topic_name());
+  std::unordered_map<std::string, int> topics_to_regularize;
+  if (!use_topic_pairs) {
+    for (auto& s : (config_.topic_name().size() ? config_.topic_name() : p_wt.topic_name())) {
+      for (int i = 0; i < p_wt.topic_name().size(); ++i) {
+        if (p_wt.topic_name(i) == s)
+          topics_to_regularize.insert(std::make_pair(s, i));
+        else
+          LOG(WARNING) << "Topic name " << s << " is not presented into model and will be ignored";
```
JeanPaulShapo: Is it ok that this warning appears for all indices `i` before break for each topic_name `s`?

PR #802, opened 2017-05-13
Fix test for DecorrelatorPhi regularizer

JeanPaulShapo:
As far as I know, we can't guarantee order of rows & columns
in `pandas.DataFrame`, which represents our Phi matrix.
That's why, in order to make proper comparison, I decided
to create `pandas.DataFrame` for reference data.


JeanPaulShapo:
@MelLain @ofrei Is it OK to use `pandas.DataFrame` in unittests?

Issue #804, opened 2017-05-19
Error in model.transform()  when using batch_vectorizer placed in memory


Angriff:
I've created BatchVectorizer in memory like this:
`bv = artm.BatchVectorizer(data_format='batches', process_in_memory=True, 
model=model, batches=batches) `

`model.fit_offline(bv)` is working good, but `model.transform(bv)` produces the error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-431-e87f49ffcd94> in <module>()
----> 1 model.transform(bv)

/home/alex/apps/anaconda2/lib/python2.7/site-packages/bigartm-0.8.3-py2.7.egg/artm/artm_model.pyc in transform(self, batch_vectorizer, theta_matrix_type, predict_class_id)
    930             theta_matrix_type_real = const.ThetaMatrixType_Cache
    931 
--> 932         batches_list = [batch.filename for batch in batch_vectorizer.batches_list]
    933 
    934         theta_info, numpy_ndarray = self._wait_for_batches_processed(

AttributeError: 'unicode' object has no attribute 'filename'
```
The workaround for me now is to go *deeper* and use 
`theta_info, numpy_ndarray = model.master.transform(batches=batches)`

PR #806, opened 2017-05-30
Fix bugs in in_memory batches processing

MelLain:


MelLain:
Fix releated to issue-804

JeanPaulShapo:
@MelLain Just use sharp sign (#) before issue number to link it :)

python/artm/wrapper/spec.py
```diff
@@ -94,7 +95,7 @@ def __init__(self, name, arguments, result=None, request=None):
     ),
     CallSpec(
         'ArtmDisposeBatch',
-        [('master_id', int), ('name', str)],
+        [('master_id', int), ('name', unicode if sys.version_info.major == 2 else str)],
     ),
```
JeanPaulShapo: I think it's better to use `six` library here (i.e [text_type](https://pythonhosted.org/six/index.html#six.text_type)).

python/tests/artm/test_batches_utils.py
```diff
@@ -55,6 +55,7 @@ def test_func():
         model.fit_online(update_every=1, batch_vectorizer=in_memory_batch_vectorizer)
         assert len(model.score_tracker['perplexity'].value) == 14
 
+        del in_memory_batch_vectorizer
```
JeanPaulShapo: Do you think it's better to comment what functionality is supposed to test in this line?

python/artm/artm_model.py
```diff
@@ -929,7 +929,9 @@ def transform(self, batch_vectorizer=None, theta_matrix_type='dense_theta',
         elif theta_matrix_type == 'cache':
             theta_matrix_type_real = const.ThetaMatrixType_Cache
 
-        batches_list = [batch.filename for batch in batch_vectorizer.batches_list]
+        batches_list = batch_vectorizer.batches_list
+        if not batch_vectorizer.process_in_memory:
+            batches_list = [batch.filename for batch in batch_vectorizer.batches_list]
```
JeanPaulShapo: Looks good! 👍
Anyway, I have an idea about adding property in `BatchVectorizer` class which returns filenames or ids according to type of processing. Can it be better to hide `process_in_memory` state from external environment?

python/artm/batches_utils.py
```diff
@@ -120,15 +120,15 @@ def __init__(self, batches=None, collection_name=None, data_path='', data_format
...
-
         if self._process_in_memory:
             for batch in self._batches_list:
                 self._model.master.remove_batch(batch)
         self._process_in_memory = False
 
+        if self._remove_batches:
+            shutil.rmtree(self._target_folder)
+        self._remove_batches = False
+
```
JeanPaulShapo: Could you give me some hints why order matters?

python/artm/master_component.py
```diff
@@ -888,7 +888,8 @@ def import_batches(self, batches=None):
 
     def remove_batch(self, batch_id=None):
         """
-        :param str batches: id of batch, loaded in RAM
+        :param batch_id: id of batch, loaded in RAM
+        :type batch_id: str or unicode
         """
         if batch_id is not None:
-            self._lib.ArtmDisposeBatch(self.master_id, batch_id)
+            self._lib.ArtmDisposeBatch(self.master_id, u'{}'.format(batch_id))
```
JeanPaulShapo: There are some problems with converting unusual str-like objects to unicode in Python.
E.g. try to execute the following line:
`
unicode_sample = u'{}'.format('пример')
`

JeanPaulShapo: It may be better to accept only unicode-like objects. Or not (I don't know)?

MelLain:
thanks, I didn't know, how to link to issue)

#806 
#804 

no, it is a full fix that can be merged with closing issue.

JeanPaulShapo:
Ok, I see.

python/artm/wrapper/spec.py
MelLain: ok, I'll try it

python/tests/artm/test_batches_utils.py
MelLain: No, bv evidently should be deletable

python/artm/artm_model.py
MelLain: Maybe, I'll think about it

python/artm/batches_utils.py
MelLain: It more convenient for me, thats the only reason)

python/tests/artm/test_batches_utils.py
JeanPaulShapo: Sorry, it was ambiguous question :)
I'd like to say 'Is it better to add comment before this line explaining what you want to test?' :)

python/artm/master_component.py
MelLain: You're right, it would be better to use only unicode

python/artm/batches_utils.py
JeanPaulShapo: Ok, no problem :)

python/tests/artm/test_batches_utils.py
MelLain: No, I don't think that it's necessary)

JeanPaulShapo: ok

JeanPaulShapo:
Feel free to merge after passing CI tests :)

JeanPaulShapo:
@MelLain There are some problems with PEP8 in Travis CI:
`python/artm/artm_model.py:926:121: E501 line too long (121 > 120 characters)`

python/artm/batches_utils.py
```diff
@@ -276,9 +276,13 @@ def __reset_batch():
     @property
     def batches_list(self):
         """
-        :return: list of batches names
```
JeanPaulShapo: @MelLain Unfortunately, right now changing logic of `batches_list` property breaks code which uses it rather weird way, e.g. `hARTM` :( In my opinion, right now it's ok to use simple hack with `process_in_memory` until we understand problem with `hARTM` class

AlexeySanko:
Hi.
I understand that I'm not in time, but while changes into memory processing are in progress:
Why did you use additionally `process_in_memory` argument for BatchVectorizer.__init__?
Common way for this function as I can see is data_format usage and we can just add new one - "memory_batch".
In this case we will avoid different type comparisons into conditions like:
[https://github.com/bigartm/bigartm/pull/806/files#diff-8ac32600e1bf043123411a7dc9657e48R89](elif data_format == 'batches' or self._process_in_memory:)
Thanks.

MelLain:
The problem with hartm is that we need to deal with batches as objects in it, so I decided to separate batches objects and their identifiers.

MelLain:
@AlexeySanko 

`or self._process_in_memory:` - this check is useless, I removed it.

Also I removed `process_in_memory` flag, because it's enough to set the `model` parameter of BV to use it in "in memory" mode.

The reason I separated data_format and proc in mem flag is that proc in memory not a format, it's only the way to deal with given batches (currently from disk).

@JeanPaulShapo please, make a full review of all changes again.

JeanPaulShapo:
Hi, @AlexeySanko.
Your suggestion looks good.
However, proposed change is likely to be backward-incompatible, because you want to remove some existing functionality from API.
I think it'll be better if you open issue about this change, and we consider it when making new major version.

MelLain:
It is not a large api breaking change, so it can be safely pushed in new release.

JeanPaulShapo:
@MelLain
OK, I agree.
Is it better to add some description about it in Release Notes (in this PR, apparently)?

MelLain:
Yes, we only need to say, that in memory flag is deprecated and should be simply deleted in all code cases.

JeanPaulShapo:
@MelLain Great!

AlexeySanko:
Hi, @MelLain.
My outside stupid view says that "batch from disk" and "batch from memory" are different formats. :) Because first are just files and second - Python list (I can create it manually without any file usage). I understand that we get the same batch list finally, but now we're talking about third-party user who doesn't know.

Now there is hidden logic: `self._process_in_memory = data_format == 'batches' and model is not None`
And arguments list do not provide clear understanding how to force in-memory processing.
Rename `model` argument to `process_in_memory_model`?
Add "in_memory_batch" data_format and make first row data_format changing for keeping similar processing for in-memory and not?
`if data_format == "in_memory_batch":
   self._process_in_memory = True
   data_format = "batch"
`

Doesn't look like very beautiful, but will make user's life easier.

Thanks.

MelLain:
@AlexeySanko I think, that `process_in_memory_model` renaming is the best solution)

AlexeySanko:
@MelLain, ok. Thanks!

python/artm/batches_utils.py
```diff
@@ -73,17 +72,18 @@ def __init__(self, batches=None, collection_name=None, data_path='', data_format
...
+                                                  disk in RAM (only if data_format == Batches).\
+                                                  NOTE: makes vectorizer model specific.
         """
         self._remove_batches = False
-        self._process_in_memory = process_in_memory and data_format == 'batches' and model is not None
-        if self._process_in_memory != process_in_memory:
-            raise IOError("Correct configuration: process_memory == True + data_format == 'batches' + model != None")
+        self._process_in_memory = data_format == 'batches' and process_in_memory_model is not None
+        if not self._process_in_memory and process_in_memory_model is not None:
+            raise IOError("Correct configuration for in memory processing: data_format ==" +\
```
AlexeySanko: Into brackets Python provides string concatenation without + and \
```
            raise IOError("Correct configuration for in memory processing: data_format =="
                          "'batches' + process_in_memory_model != None")
```

docs/release_notes/python.txt
```diff
@@ -8,6 +8,10 @@ For this reason we recommend to review :doc:`messages`.
 For further reference about Python API refer to :doc:`/api_references/python_interface/artm_model`,
 `Q & A <https://github.com/bigartm/bigartm/wiki/Q&A>`_ or `tutorials <https://github.com/bigartm/bigartm-book/blob/master/README.md>`_.
 
+v0.8.4
+------
+* Flag ``process_in_memory`` was removed from ``artm.BatchVectorizer``, and ``model`` parameter ws renames into ``process_in_memory_model``. Now specification of this parameter with ``data_format='batches'`` will mean in memory processing of given in ``batches`` parameter batches.
```
JeanPaulShapo: Try to write in *imperative mood* like `Remove flag ... from ..., and rename ... to ...`

JeanPaulShapo: *in memory* -> *in-memory* (example *in-memory database*)

JeanPaulShapo: I think, it'll be better to change *processing of given in ``batches`` parameter batches* to *processing of batches, specified in ``batches`` parameter*

JeanPaulShapo: *Now specification of this parameter with ... will mean ...* -> *Passing this argument with ... triggers ...*

python/artm/artm_model.py
```diff
@@ -582,10 +574,10 @@ def _func(batch):
 
         self._wait_for_batches_processed(
             self._pool.apply_async(func=self.master.fit_online,
-                                   args=(batches_list, batch_vectorizer.weights,
+                                   args=(batch_vectorizer.batches_ids, batch_vectorizer.weights,
                                          update_after_final, apply_weight_final,
                                          decay_weight_final, async)),
-            len(batches_list))
+            len(batch_vectorizer.batches_ids))
```
JeanPaulShapo: Is it OK to write `__len__` function to count number of entities in `batch_vectorizer` instead `num_batches` property and even ugly `len(batch_vectorizer.batches_ids)`, like in file `Scores` class from `scores.py`?

python/artm/batches_utils.py
```diff
@@ -73,17 +72,18 @@ def __init__(self, batches=None, collection_name=None, data_path='', data_format
                                        and if data_weight is list - automatically set to False
         :param class_ids: list of class_ids or single class_id to parse and include in batches
         :type class_ids: list of str or str
-        :param bool process_in_memory: process batches from disk or from RAM \
-                                       (only if data_format == Batches), model parameter is required
-        :param artm.ARTM model: ARTM instance that will use this vectorizer, is required when\
-                                process_in_memory == True. Will be ignored in other cases
+        :param artm.ARTM process_in_memory_model: ARTM instance that will use this vectorizer, is\
+                                                  required when one need processing of batches from\
```
JeanPaulShapo: *one need**s** *?

python/artm/batches_utils.py
```diff
@@ -73,17 +72,18 @@ def __init__(self, batches=None, collection_name=None, data_path='', data_format
...
+                                                  disk in RAM (only if data_format == Batches).\
+                                                  NOTE: makes vectorizer model specific.
```
JeanPaulShapo: lower `Batches` and enclose it in quotes

python/artm/artm_model.py
MelLain: I agree, that len(...) should be replaces with method call, but I don't think, that `__len__` is a good idea, because `num_batches` is more convenient.

JeanPaulShapo:
Fixes #804

PR #814, opened 2017-06-12
Perplexity refactoring and fixing

MelLain:


src/artm/score/perplexity.cc
```diff
@@ -33,52 +33,66 @@ void Perplexity::AppendScore(
...
+          class_weight_map.insert(std::make_pair(args.class_id(i), args.class_weight(i)));
+          normalizer_map.insert(std::make_pair(args.class_id(i), 0.0));
+          raw_map.insert(std::make_pair(args.class_id(i), 0.0));
+          zero_words_map.insert(std::make_pair(args.class_id(i), 0));
           break;
         }
+      }
+    }
+    if (class_weight_map.empty()) {
+      LOG(ERROR) << "All requested class ids are not presented in model. Score calculation will be skipped";
```
JeanPaulShapo: @MelLain Is it better to write `None of requested classes were presented in model. ...`?

src/artm/score/perplexity.cc
```diff
@@ -163,17 +191,70 @@ void Perplexity::AppendScore(const Score& score, Score* target) {
...
+        dst->set_normalizer(src.normalizer());
+        dst->set_raw(src.raw());
+        dst->set_zero_words(src.zero_words());
+      }
+    }
+
+    for (size_t j = 0; j < perplexity_target->class_id_info_size(); ++j) {
+      auto score = perplexity_target->class_id_info(j);
+      pre_value += score.raw() / score.normalizer();
+
```
JeanPaulShapo: @MelLain Should we weigh log-likelihoods of different modalities according to their `class_weight` values?

python/artm/score_tracker.py
```diff
@@ -128,16 +135,21 @@ def __init__(self, score):
...
         * Note: every field is a list of info about score on all synchronizations.
         * value - values of perplexity.
-        * raw - raw values in formula for perplexity.
-        * normalizer - normalizer values in formula for perplexity.
-        * zero_tokens - number of zero p(w|d) = sum_t p(w|t) p(t|d).
+        * raw - raw values in formula for perplexity (in case of one class id).
+        * normalizer - normalizer values in formula for perplexity  (in case of one class id).
+        * zero_tokens - number of zero p(w|d) = sum_t p(w|t) p(t|d)  (in case of one class id).
+        * class_id_info - array of structures, each structure contains raw, normalizer\
+                          zero_tokens and class_id name (in case of several class ids).
```
MelLain: I saved them not for backward compability. They are using in case of encounting all class_ids in score, because I don't want to create smth like '@all_classes' modality and put it into new filelds.

The other way is to use one ClassInfo container without class_id, but I also din't like it because it' unobvious.

src/artm/core/dictionary_operations.cc
```diff
@@ -279,15 +279,23 @@ std::shared_ptr<Dictionary> DictionaryOperations::Gather(const GatherDictionaryA
...
     for (int index = 0; index < batch.token_size(); ++index) {
       // unordered_map.operator[] creates element using default constructor if the key doesn't exist
-      TokenValues& token_info = token_freq_map[Token(batch.class_id(index), batch.token(index))];
+      ClassId token_class_id = batch.class_id(index);
+      TokenValues& token_info = token_freq_map[Token(token_class_id, batch.token(index))];
       token_info.token_tf += token_n_w[index];
-      sum_w_tf += token_n_w[index];
       token_info.token_df += token_df[index];
+
+      auto class_id_iter = sum_w_tf.find(token_class_id);
```
MelLain: I know this feature, but this place confused me:( Thanks!

src/artm/core/dictionary_operations.cc
```diff
@@ -279,15 +279,23 @@ std::shared_ptr<Dictionary> DictionaryOperations::Gather(const GatherDictionaryA
...
+        sum_w_tf[token_class_id] += token_n_w[index];
+      } else {
+        sum_w_tf[token_class_id] = token_n_w[index];
+      }
     }
   }
 
-  for (auto iter = token_freq_map.begin(); iter != token_freq_map.end(); ++iter)
-    iter->second.token_value = static_cast<float>(iter->second.token_tf / sum_w_tf);
+  for (auto iter = token_freq_map.begin(); iter != token_freq_map.end(); ++iter) {
```
MelLain: ok

src/artm_tests/test_mother.cc
```diff
@@ -12,6 +12,59 @@
...
+  }
+
+  for (int iDoc = 0; iDoc < nDocs; iDoc++) {
+    artm::Item* item = batch.add_item();
+    item->set_id(iDoc);
+    for (int iToken = 0; iToken < nTokens; ++iToken) {
+      item->add_token_id(iToken);
+      int background_count = (iToken > 40) ? (1 + rand() % 5) : 0;  // NOLINT
+      int topical_count = ((iToken < 40) && ((iToken % 10) == (iDoc % 10))) ? 10 : 0;
+      item->add_token_weight(static_cast<float>(background_count + topical_count));
```
MelLain: yes, it is the troubles of the user

src/artm_tests/multiple_classes_test.cc
```diff
@@ -420,7 +372,7 @@ TEST(MultipleClasses, WithoutDefaultClass) {
   gs.set_score_name("tts_class_one"); EXPECT_TRUE(master2.GetScoreAs< ::artm::TopTokensScore>(gs).num_entries() > 0);
   gs.set_score_name("tts_class_two"); EXPECT_TRUE(master2.GetScoreAs< ::artm::TopTokensScore>(gs).num_entries() > 0);
 
-  gs.set_score_name("perplexity");
+  gs.set_score_name("perplexity"); auto t = master.GetScoreAs< ::artm::PerplexityScore>(gs);
```
MelLain: I forgot to remove it after testing(

python/tests/wrapper/test_02_regularizers.py
```diff
@@ -33,7 +33,7 @@ def test_func():
     num_document_passes = 10
     num_outer_iterations = 8
 
-    perplexity_tol = 0.001
+    perplexity_tol = 1.0
```
MelLain: Test on my pc gives result with little difference compared to results on CI. It's normal to compare perplexities in range 1-10.

python/artm/score_tracker.py
JeanPaulShapo: What about excluding `class_id` from `ClassInfo` containter and add `repeated class_id` into `PerplexityScore`? In that case we can reuse the resulting container to show score for all class-ids.

utils/cpplint_files.txt
```diff
@@ -50,6 +50,7 @@ src/artm_tests/collection_parser_test.cc
 src/artm_tests/cpp_interface_test.cc
 src/artm_tests/template_manager_test.cc
 src/artm_tests/regularizers_test.cc
+src/artm_tests/scores_test.cc
 src/artm_tests/repeatable_result_test.cc
```
MelLain: there is no alphabetic order, so in this file it can be skipped

python/artm/score_tracker.py
MelLain: This is a good idea, but let's stay it for next review, I'll fix it.

src/artm/score/perplexity.cc
MelLain: No, class_weigh is taken into account in this line (~120)

float token_weight = class_weight * item.token_weight(token_index);

python/artm/score_tracker.py
JeanPaulShapo: Agree

src/artm/score/perplexity.cc
JeanPaulShapo: But you also multiply normalizer by the same value in line 82.
Therefore, `class_weight` is present both in numerator (`score.raw()`) and denominator (`score.normalizer()`).
Am I right or am I missing something again?But you also multiply normalizer by the same value in line 82.
Therefore, `class_weight` is present both in numerator (`score.raw()`) and denominator (`score.normalizer()`).
Am I right or am I missing something again?

utils/cpplint_files.txt
JeanPaulShapo: Ahhh, I see)

src/artm/score/perplexity.cc
MelLain: You're right, and this is a correct behaviour - using modalities in fact means multipling token weight on class weight, and the same token weight should be both in raw and normilazer, as I think

src/artm/core/dictionary_operations.cc
JeanPaulShapo: @MelLain I think you forgot to remove initial code fix :)

src/artm/score/perplexity.cc
JeanPaulShapo: @MelLain So I think I don't understand the multimodal version of perplexity: why we must sum normalized log-likelihoods for all modalities with equal weights, but not with weights, proportional to `class_weight` values? Is there any whitepaper with direct formula?

MelLain: I don't understand you( We count perplexity separately for each modaility and everywhere use tokens weights multiplied on corresponding modality weight

JeanPaulShapo:
@MelLain 👍 Feel free to merge after passing CI. However, don't forget to write 'good' message of merge commit :)

PR #816, opened 2017-06-24
Model Saving From Python API

MelLain:
This change will be done step by step with CI checking of each part:

1) prepare cpp part
- add ability to save and load score tracker
- tests
2) python API part
- join saving and loading of score tracker and phi matricies (including theta) in one method and one function
- tests
- docs and examples

### DIFF
--- a/python/artm/artm_model.py
+++ b/python/artm/artm_model.py
@@ -146,7 +146,8 @@ class ARTM(object):
             Later you can retrieve this matix with ARTM.get_phi(model_name=ARTM.theta_name),\
             change its values with ARTM.master.attach_model(model=ARTM.theta_name),\
             export/import this matrix with ARTM.master.export_model('ptd', filename) and\
-            ARTM.master.import_model('ptd', file_name).
+            ARTM.master.import_model('ptd', file_name). In this case you also are able to work\
+            with theta matrix when using 'dump_model' method and 'load_model' function.
         """
         self._num_processors = None
         self._cache_theta = False
@@ -1011,6 +1012,17 @@ class ARTM(object):
         return 'artm.ARTM(num_topics={0}, num_tokens={1}{2})'.format(
             self.num_topics, num_tokens, class_ids)
 
+    def dump_model(self, target_folder):
+        """
+        descr
+        """
+        pass
 
 def version():
     return ARTM(num_topics=1).library_version
+
+def load_model(target_folder):
+    """
+    descr
+    """
+    return ARTM(num_topics=1)
--- a/python/artm/artm_model.py
+++ b/python/artm/artm_model.py
@@ -1088,7 +1088,7 @@ class ARTM(object):
         with open(os.path.join(data_path, PARAMETERS_FILENAME_JSON), 'w') as fout:
             json.dump(params, fout)
 
-        # save parameters in humanreadable format
+        # save parameters in binary format
         regularizers = {}
         for name, regularizer in iteritems(self._regularizers._data):
             regularizers[name] = [regularizer._config_message.__name__,
@@ -1097,15 +1097,15 @@ class ARTM(object):
             tau = None
             gamma = None
             try:
-                tau = score.tau
-                gamma = score.gamma
+                tau = regularizer.tau
+                gamma = regularizer.gamma
             except:
                 pass
 
             if tau is not None:
-                scores[name].append(tau)
+                regularizers[name].append(tau)
                 if gamma is not None:
-                    scores[name].append(gamma)
+                    regularizers[name].append(gamma)
 
         params['regularizers'] = regularizers
 
--- a/python/artm/artm_model.py
+++ b/python/artm/artm_model.py
@@ -7,6 +7,9 @@ import glob
 import shutil
 import tempfile
 import numpy
+import datetime
+import json
+import pickle
 
 from pandas import DataFrame
 from six import iteritems, string_types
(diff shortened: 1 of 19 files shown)


src/artm/core/master_component.cc
```diff
@@ -412,10 +412,74 @@ void MasterComponent::ImportModel(const ImportModelArgs& args) {
...
+
+  LOG(INFO) << "Exporting score tracker to " << args.file_name();
+
+  const char version = 0;
+  fout << version;
+
+  // We expect here that each ScoreData object has suitable size (< 2GB)
+  for (auto& item : instance_->score_tracker()->GetDataUnsafe()) {
+    auto str = item->SerializeAsString();
+    fout << str.size();
```
MelLain: You're right, but this is really a common troubles of our self-written code. I believe, that it will be replaces with some standart tools in future

python/artm/artm_model.py
```diff
@@ -1011,6 +1023,185 @@ def __repr__(self):
...
+
+        with open(os.path.join(data_path, PARAMETERS_FILENAME_BIN), 'wb') as fout:
+            pickle.dump(params, fout)
+
 
 def version():
     return ARTM(num_topics=1).library_version
+
+
+def load_artm_model(data_path):
```
MelLain: I wanted to make save/load in the same way, as it done in pandas:

pandas.DataFrame.to_csv
pandas.read_csv

As I know, they didn't use any modifier to make this function a method.

src/artm/core/master_component.cc
JeanPaulShapo: Ok, then it'll be better to somehow track these pieces of ... self-written code.

python/artm/artm_model.py
JeanPaulShapo: OK, it seems reasonable. By the way, there are plenty of choices:

* The aforementioned `read_csv` method can return not only `DataFrame`, but also `TextParser` (i.e. some kind of iterator);
* `numpy` library uses functions, not methods, for both loading and saving, and in both cases it operates on array;
* `scipy` uses the same policy for sparse matrices as `numpy`; in their case, there are more than one possible type of argument;

MelLain: Of course, I agree:) I only chose this one)

python/artm/artm_model.py
```diff
@@ -1011,6 +1023,185 @@ def __repr__(self):
...
+        params['synchronizations_processed'] = self._synchronizations_processed
+        params['num_online_processed_batches'] = self._num_online_processed_batches
+        params['initialized'] = self._initialized
+
+        regularizers = {}
+        for name, regularizer in iteritems(self._regularizers.data):
+            tau = None
+            gamma = None
+            try:
+                tau = score.tau
```
MelLain: bug. iteration over regularizers, but using scores

MelLain: 1) fix it
2) write test for human-readable part.
3) catch only specific exception in all try-catch blocks (10 places)

JeanPaulShapo: Good catch!

JeanPaulShapo: And that's why we have to catch only specific classes :)

python/artm/artm_model.py
```diff
@@ -1067,9 +1067,9 @@ def dump_artm_model(self, data_path):
             tau = None
             gamma = None
             try:
-                tau = score.tau
-                gamma = score.gamma
-            except:
+                tau = regularizer.tau
+                gamma = regularizer.gamma
+            except KeyError:
```
JeanPaulShapo: There is no need in try-except clause.

python/artm/artm_model.py
```diff
@@ -1099,7 +1099,7 @@ def dump_artm_model(self, data_path):
             try:
                 tau = regularizer.tau
                 gamma = regularizer.gamma
-            except:
+            except KeyError:
```
JeanPaulShapo: There is no need in try-except clause.

python/artm/regularizers.py
```diff
@@ -232,7 +232,7 @@ def __init__(self, name, tau, gamma, config, topic_names, class_ids, dictionary)
             try:
                 if config.HasField('dictionary_name'):
                     self._dictionary_name = config.dictionary_name
-            except:
+            except ValueError:
```
JeanPaulShapo: I suggest to rewrite such try-except clause, because `HasField` method works not as someone can expect: quote from [StackOverflow](https://stackoverflow.com/a/36742793)

"I think you've misunderstood what HasField() does. It does not check whether a protobuf type defines a specific field by a name. What it does is to check if, given the name of a message field, that that field is set for the current instance."

My suggestion:
`
try:
    dict_name = config.dictionary_name
    self._dictionary_name = config.dictionary_name
except AttributeError:
    pass
`

python/artm/scores.py
```diff
@@ -110,7 +110,7 @@ def __init__(self, name, class_id, topic_names, model_name, config):
             try:
                 if config.HasField('class_id'):
                     self._class_id = config.class_id
-            except:
+            except ValueError:
```
JeanPaulShapo: I suggest to rewrite such try-except clause, because `HasField` method works not as someone can expect: quote from [StackOverflow](https://stackoverflow.com/a/36742793)

"I think you've misunderstood what HasField() does. It does not check whether a protobuf type defines a specific field by a name. What it does is to check if, given the name of a message field, that that field is set for the current instance."

My suggestion:


`try:`
`    self._class_id = config.class_id`
`except AttributeError:`
`    pass`

JeanPaulShapo:
:+1: Feel free to merge after CI tests

PR #818, opened 2017-07-18
add new time regularizer

MelLain:


src/artm/regularizer/smooth_time_in_topics_phi.h
```diff
@@ -0,0 +1,52 @@
...
+
+   Author: Murat Apishev (great-mel@yandex.ru)
+
+   This class proceeds smoothing of tokens in Phi using nearest values.
+
+   More preferable for time stamps tokens. Requires to be used with sorted dictionary
+   (e.g. the tokens should follow in some order, chronological, for instance).
+
+   The formula of M-step is
+
```
MelLain: Oh, you are right. I will resolve it in next way:
1) class_ids -> class_id -- I don't think, that using this regularizer with several modalities is correct
2) we suppose, that tokens of one modality are sorted, but we should hardcode restriction, that the are going as one block (e.g. that should be allowed to intersect with tokens of other modalities). To deal with it we need a minor modification of current regularization loop.

src/artm/regularizer/smooth_time_in_topics_phi.cc
```diff
@@ -25,27 +25,38 @@ bool SmoothTimeInTopicsPhi::RegularizePhi(const ::artm::core::PhiMatrix& p_wt,
...
+    }
 
     for (int topic_id = 0; topic_id < topic_size; ++topic_id) {
       if (topics_to_regularize[topic_id]) {
-        double value = p_wt.get(token_id, topic_id);
+        double value = p_wt.get(index_prev, topic_id);
 
-        value *= ((p_wt.get(token_id - 1, topic_id) - value) > 0.0 ? 1.0 : -1.0) +
-                 ((p_wt.get(token_id + 1, topic_id) - value) > 0.0 ? 1.0 : -1.0);
+        value *= ((p_wt.get(index_prev_prev, index_prev) - value) > 0.0 ? 1.0 : -1.0) +
```
JeanPaulShapo: I suppose there must be `topic_id` somewhere in `p_wt.get` call)

JeanPaulShapo: Maybe instead of `index_prev`

PR #824, opened 2017-07-26
Refactoring

MelLain:
List of changes:

- all unnecessary doubles were replaced with floats
- everywhere if and for statements are using with brackets
- use const qualifiers for references in range based for
- replace guard headers with \#pragma once (update cpplint)
- some code decoration in regularizers and scores

According to these changes let's fix some codestyle rules:

1) use float as floating point type everywhere, where you can do it. Each usage of double type should have reasons (we don't use long instead of int everywhere, yes?)

2) each new header should have \#pragma once in the begining

3) Use const qualifier where it is possible, especially in range-based for

4) body of every if-for-while statement should be bordered with brackets { }, no matter the body contains one line or more.

5) Empty {} should be written with space (e.g. { })

6) we use Constructor() : param_(param) only in case of single parameter or short list of parameters with empty initialisation (: a_(), b_()). In other cases type

Constructor()
    : param_(param)
    , param2_(param2_)

the indention is 4 spaces before ":" and ","

7) Try to write 'f' suffix when using float numbers (e.g. 1.0f)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,8 @@ src/Win32/*
 *.log.ERROR.*
 *.batch
 
+*\#*\#*
+
 # Binaries
 /src/artm/unittests/
 /src/cpp_client/cpp_client
--- a/python/artm/master_component.py
+++ b/python/artm/master_component.py
@@ -513,8 +513,8 @@ class MasterComponent(object):
         :param regularizer_name: list of names of Phi regularizers to use
         :type regularizer_name: list of str
         :param regularizer_tau: list of tau coefficients for Phi regularizers
-        :type regularizer_tau: list of double
-        :type regularizer_tau: list of double
+        :type regularizer_tau: list of floats
+        :type regularizer_tau: list of floats
         """
         args = messages.RegularizeModelArgs(pwt_source_name=pwt,
                                             nwt_source_name=nwt,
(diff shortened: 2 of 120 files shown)


ofrei:
@MelLain Please give me time until tomorrow and I'll review, thanks! :)

ofrei:
I agree with most changes, and overall this change definitely improve the code -- please ahead and merge it as it is (so don't roll back changes based on my other comments).

ofrei:
Few comments with references to https://google.github.io/styleguide/cppguide.html
>Short conditional statements may be written on one line if this enhances readability. You may use this only when the line is brief and the statement does not use the else clause.
Example: code in ``c_interface.cc``, method ``static void EnableLogging(artm::ConfigureLoggingArgs* args)`` -- many short if statements after ``  // Setting all other flags except log_dir``.

ofrei:
https://google.github.io/styleguide/cppguide.html#Windows_Code
>Do not use #pragma once; instead use the standard Google include guards

Good discussion here:
https://stackoverflow.com/questions/787533/is-pragma-once-a-safe-include-guard

So there are pros and cons, but for BigARTM code I'm fine with changing everything to ``#pragma once``.

ofrei:
double vs float --- this is a difficult question
https://stackoverflow.com/questions/1074474/should-i-use-double-or-float
Some cases in bigartm require float because it saves memory. Other places like perplexity calculation it is nice to have double.  I'm OK with float's everywhere (except perplexity) -- let's keep it this way and see if anything bad ever happens.

ofrei:
I know @sashafrey won't object , just assume he approved this too. 👍

ofrei:
Ah, sorry - merge conflict is because I've changed readme in the root with the link to [codestyle wiki](https://github.com/bigartm/bigartm/wiki/Code-style). think it will be nice to remove code style from sphynx documentation - it is too hard to update it there

PR #826, opened 2017-07-29
Dictionary fix and improvment

MelLain:


python/artm/dictionary.py
```diff
@@ -106,7 +107,9 @@ def load_text(self, dictionary_path, encoding='utf-8'):
         self._reset()
         dictionary_data = messages.DictionaryData()
         with codecs.open(dictionary_path, 'r', encoding) as fin:
-            dictionary_data.name = fin.readline().split(' ')[1][0: -1]
+            first_str = fin.readline()[: -1].split(' ')
```
MelLain: Here we made old text dictionaries incompatible, but I don't think that it's very bad:

1) people mostly store dictionaries in binary format, text is smth temporary (binary format exporting didn't change, only import was fixed)
2) text dictionary can be simply updated by hand (by appending " num_items: <num_items>" to the first line)

python/artm/dictionary.py
```diff
@@ -154,7 +157,7 @@ def gather(self, data_path, cooc_file_path=None, vocab_file_path=None, symmetric
                                        symmetric_cooc_values=symmetric_cooc_values)
 
     def filter(self, class_id=None, min_df=None, max_df=None, min_df_rate=None, max_df_rate=None,
-               min_tf=None, max_tf=None, max_dictionary_size=None):
+               min_tf=None, max_tf=None, max_dictionary_size=None, recalculate_value=False):
```
MelLain: Note: recalculate_value == True is useful in dictionary for perplexity calculation, False - for dictionary with custom 'value' values for regularization

PR #830, opened 2017-08-04
Build standalone Python wheels

vmarkovtsev:
I call "a standalone wheel" a package which does not require the libartm installation - aka "tensorflow style". Those wheels can be safely pushed on PyPi and will just work.

I had to change:

* `python/CMakeLists.txt`: add another custom target to execute ` setup.py bdist_wheel`
* `python/setup.py`: hack setuptools to include the needed shared library
* `python/artm/wrapper/api.py`: refactor `_load_cdll()` to load the library from several places - the common practice. The env variable is the most important, then the default relative path and finally the absolute path in the package root.
* docs - as far as I could

Besides, there seems to be a bug in TravisCI configuration regarding python command detection. There is no explicit `-DPYTHON` in the build script, and by default cmake chooses `python2` which prevents from testing py3.x:

```
if (MSVC OR APPLE)
set(PYTHON python CACHE INTERNAL "Python command")
else (MSVC OR APPLE)
set(PYTHON python2 CACHE INTERNAL "Python command")
endif (MSVC OR APPLE)
```

`bdist_wheel` command helped to find this bug: there is no `wheel` package installed in py2 env in Travis.
I dared to remove `python2` branch completely.

### DIFF
--- a/python/artm/wrapper/api.py
+++ b/python/artm/wrapper/api.py
@@ -42,22 +42,35 @@ class LibArtm(object):
         if sys.platform.startswith('darwin'):
             default_lib_name = 'libartm.dylib'
 
-        if lib_name is None:
-            # try to get library path from environment variable
-            lib_name = os.environ.get('ARTM_SHARED_LIBRARY')
-
-        if lib_name is None:
-            # set the default library name
-            lib_name = default_lib_name
-
-        try:
-            cdll = ctypes.CDLL(lib_name)
-        except OSError as e:
+        lib_names = []
+        
+        if lib_name is not None:
+            lib_names.append(lib_name)
+            
+        env_lib_name = os.environ.get('ARTM_SHARED_LIBRARY')
+        if env_lib_name is not None:
+            lib_names.append(env_lib_name)
+        
+        lib_names.append(default_lib_name)
+        lib_names.append(os.path.join(os.path.dirname(__file__), "..", default_lib_name))
+        
+        # We look into 4 places: lib_name, ARTM_SHARED_LIBRARY, default_lib_name and
+        # default_lib_name in the python package root
+        cdll = None
+        for ln in lib_names:
+            try:
+                cdll = ctypes.CDLL(ln)
+                break
+            except OSError as e:
+                if ln == default_lib_name:
+                    exc = e
+                continue
+        if cdll is None:
             exception_message = (
-                '{e}\n'
-                'Failed to load artm shared library. '
+                '{exc}\n'
+                'Failed to load artm shared library from `{lib_names}`. '
                 'Try to add the location of `{default_lib_name}` file into your PATH '
-                'system variable, or to set ARTM_SHARED_LIBRARY - a specific system variable '
+                'system variable, or to set ARTM_SHARED_LIBRARY - the specific system variable '
                 'which may point to `{default_lib_name}` file, including the full path.'
             ).format(**locals())
             raise OSError(exception_message)
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ env:
 before_install:
   # we need latest pip to work with only-binary option
   - pip install -U pip
-  - pip install -U pytest pep8
+  - pip install -U pytest pep8 wheel
   - pip install -U numpy pandas tqdm --only-binary numpy pandas
   - pip install protobuf==3.0.0
   # configure ccache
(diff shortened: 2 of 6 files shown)


JeanPaulShapo:
Great job! I'll try to review this PR in the weekend.

vmarkovtsev:
Ping.

JeanPaulShapo:
SYN-ACK

JeanPaulShapo:
Could you please describe common usage pattern of this method? Should it be used **only** by skilled developers just to upload to PyPi? Also, if you have some links to useful material, it'll be great if you share it with us.

vmarkovtsev:
The usage pattern is pretty much the same as with Tensorflow: we install the wheel inside our docker environment without the need to compile the library which is so painful in that environment. It is actually easier and more reliable than installing the whole bigartm suite with CLI and `protoc` (btw, is it really needed?), and is guaranteed to work in Windows/macOS with possibly screwed library paths.

I would suggest to upload this wheel to PyPi - make a binary release after applying `auditwheel repair`. There will be no need for end users to build the package from source.

python/artm/wrapper/api.py
```diff
@@ -42,22 +42,35 @@ def _load_cdll(self, lib_name):
...
+        
+        if lib_name is not None:
+            lib_names.append(lib_name)
+            
+        env_lib_name = os.environ.get('ARTM_SHARED_LIBRARY')
+        if env_lib_name is not None:
+            lib_names.append(env_lib_name)
+        
+        lib_names.append(default_lib_name)
+        lib_names.append(os.path.join(os.path.dirname(__file__), "..", default_lib_name))
```
JeanPaulShapo: Maybe there is have to be one additional '..' folder?

JeanPaulShapo: Thanks, I got it.

JeanPaulShapo:
Generally this PR seems rather OK, however there are some few questions left:

* Should we check installation from wheels on continuous integration?
* What credentials (e.g. maintainers and email) to supply when uploading wheel to PyPI? There is an example in [the file from outdated PR](https://github.com/JeanPaulShapo/bigartm/blob/cf6773f15aecf378dea4856f156509e3b9dfb15d/setup.py). 

And sorry, I really need some time to refresh my mind on this `setuptools`-`distutils` topic.

JeanPaulShapo:
And you'd better rebase the branch.

vmarkovtsev:
* Checking the installation: yes, we should, but let's do it in a separate PR
* You are the maintainers :) Take for example [ours](https://github.com/src-d/ast2vec/blob/master/setup.py) I am not uploading anything myself anyway.

Refreshing the topic: Armin's nice blog post http://lucumr.pocoo.org/2012/6/22/hate-hate-hate-everywhere/

vmarkovtsev:
...and rebased.

JeanPaulShapo:
Thanks for the article. I'll try to respond in two days.

JeanPaulShapo:
@vmarkovtsev 

As I've got some information about BigARTM Windows distribution from @ofrei, you need not only `artm.dll` library but also Visual Studio runtime libraries (namely, `msvcp140.dll` and `vcruntime140.dll`). Therefore, in case of Windows we mustn't consider path to python package root - you need to add this path only when `'win' no in sys.platform`.
Now, consider we add check for windows/non-windows. Then, if you intend to use this code only as some kind of local Python binary package on non-windows, this PR seem OK for me. I think we can have some separate PR to fully integrate our ibrary with PyPI.

So, if I understand you correctly, you only have to rebase PR again before merging.

python/artm/wrapper/api.py
JeanPaulShapo: I think we mustn't use for Windows, i.e. include this path only conditionallt.

ofrei:
@vmarkovtsev @JeanPaulShapo Sorry for my delay. Excellent PR, thanks! I've tested that this PR doesn't break current installation steps (see [these](https://github.com/bigartm/bigartm-install/pull/5) CI runs in ``bigartm-install`` repo). Since current scenarios work we can safely merge this PR, and address the remaining questions separately (including Windows story for the bigartm wheel, putting it on PyPi, etc).

vmarkovtsev:
Yeah, sorry guys, got a bit sick lately, should have rebased this myself.

ofrei:
@vmarkovtsev no problem! if it takes 2 weeks for us to review it feels to me like our responsibility to bring in this change

JeanPaulShapo:
@vmarkovtsev Could you please add some comments about my thoughts on Windows distribution?

ofrei:
@JeanPaulShapo Regarding Windows wheel... when we produce binary package for Windows there is no problem with VS runtime libraries --- thanks to CMake ``InstallRequiredSystemLibraries`` flag. I'm not sure how to automatically include the right version of the VS runtime libraries into python wheel, because it depends on the version of visual studio that is used. I think it needs to be tested separately. For now it is perfect if we can make Wheel work on Linux.

JeanPaulShapo:
@ofrei OK, in that case there is no problem left. Feel free for merge corresponding PR after CI.

ofrei:
Merged through #838

PR #833, opened 2017-08-10
Python 3.x support for hierarchical topic models

AVBelyy:


python/artm/hierarchy_utils.py
```diff
@@ -386,14 +386,14 @@ def load(self, path):
...
-        with open(info_filename[0]) as fin:
+        with open(info_filename[0], "rb") as fin:
             info = pickle.load(fin)
         model_filenames = glob.glob(os.path.join(path, "*.model"))
         if len({len(info["parent_level_weight"]) + 1, len(model_filenames) / 2}) > 1:
             raise ValueError("Given path is not hARTM safe")
         self._levels = []
         sorted_model_filenames = sorted(model_filenames)
-        for level_idx in xrange(len(model_filenames) / 2):
+        for level_idx in range(int(len(model_filenames) / 2)):
```
JeanPaulShapo: Use integer division.

python/artm/hierarchy_utils.py
```diff
@@ -287,7 +287,7 @@ def del_level(self, level_idx):
         if level_idx == -1:
             del self._levels[-1]
             return
-        for _ in xrange(level_idx, len(self._levels)):
+        for _ in range(level_idx, len(self._levels)):
```
JeanPaulShapo: Import `range` from `six.moves` package, as in `batches_utils.py`

python/artm/hierarchy_utils.py
```diff
@@ -363,9 +363,9 @@ def save(self, path):
             level.save(os.path.join(path, "level" +
                                     str(level_idx) + "_pwt.model"), model_name="p_wt")
         info = {"parent_level_weight": [
-            level.phi_batch_weight for level in self._levels[1:]]}
+            level._phi_batch_weight for level in self._levels[1:]]}
         with open(os.path.join(path, "info.dump"), "wb") as fout:
-            pickle.dump(info, fout)
+            pickle.dump(info, fout, protocol=2)
```
JeanPaulShapo: I think it's also better to load `pickle` from `six.moves`.

python/artm/hierarchy_utils.py
```diff
@@ -363,9 +363,9 @@ def save(self, path):
             level.save(os.path.join(path, "level" +
                                     str(level_idx) + "_pwt.model"), model_name="p_wt")
         info = {"parent_level_weight": [
-            level.phi_batch_weight for level in self._levels[1:]]}
+            level._phi_batch_weight for level in self._levels[1:]]}
```
JeanPaulShapo: Use property  `parent_level_weight` instead, as we needn't rely on any attributes starting with underscore.

python/artm/hierarchy_utils.py
AVBelyy: I think, for consistency with other code, we can just use `pickle` with default protocol version.

JeanPaulShapo: Sounds reasonable

JeanPaulShapo:
Good job!

PR #854, opened 2017-11-27
Update version of cmake in docs to 3.1, minor fixes

zZLOiz:
CMake config for python wheel (`python/CMakeLists.txt`) uses `cmake -E env` command which was implemented only in CMake 3.1 (https://github.com/Kitware/CMake/commit/7abd574798f9900abfe502f3941cffaa774062b1). So some distributives (at least Centos 7), which still use CMake 2.8 as default, produce a quite ambiguous error message. PR contains two updates for docs: an update version of CMake to 3.1 and a copy note about shared versions of libraries to the linux installation page also (shared libraries are installed by default in Centos 7 and Gentoo).

### DIFF
--- a/docs/devguide/dev_build_linux.txt
+++ b/docs/devguide/dev_build_linux.txt
@@ -10,7 +10,7 @@ System dependencies
 Building BigARTM requires the following components:
 
     * `git <https://git-scm.org>`_ (any recent version) -- for obtaining source code;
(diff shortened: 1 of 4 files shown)


JeanPaulShapo:
@zZLOiz There are some problems with continous integration, I'll reopen this PR to see if it solves them.

JeanPaulShapo:
@zZLOiz Good job!

PR #856, opened 2017-12-02
Mel lain transaction artm

MelLain:
Adding basic support of transaction models


ofrei:
@MelLain how is it going with this PR? Some random thoughts from my side:
- rationale for this PR is OK. We want this feature in master asap! :) Transactions is important feature, it nicely fits to bigartm paradigm.
- we need good documentation to make sure users who already used class_id (aka modalities) know how to deal with transaction_id. It migth be somewhat anoying for those users who are 100% happy with modalities -- to help them we need good info in release notes.
- changes in python API are ok
- changes in protobuf messages are slightly concerning. If users ever saved serialized NetPlsaPhiConfig it will breack deserialization. I guess we don't care.  Are we concerned with backward compatibility in PerplexityScoreConfig , PerplexityScore  serialization? I don't remember any place in the code where we serialize them to disk. If we don't -- fine, no problem. If we do, those places must be tested (changes to those protobuf messages might in fact be backwards compatible, but if we rely on this it needs to be tested,
- skip batch dictionary test is removed. this is OK, I don't think this feature is actively used - but we need to create a github issue for this. 
- performance-wise I'm sure it is possible to implement in an efficient way in processor. Are you planing to cleanup the code? If yes, once you have a clean code, I'd be happy to do some profiling and test before/after.

ofrei:
:+1:

PR #857, opened 2017-12-04
Update dependencies, support compilation on MinGW

maksbotan:
Hi. I work on bigartm-related stuff for LINIS laboratory of HSE in Saint Petersburg. I prefer gcc (MinGW) compiler for my code, so I've added MinGW support to bigartm.

Here's a summary of my changes:

- Updated bundled gflags and glog to latest releases, which support mingw out of the box
- Extracted `DLL_PUBLIC` define to a separate header which is generated by CMake
- Used `#ifdef _MSC_VER` instead of `_WIN32` for checks concerning compiler, not platform. For example for `#pragma warning` support
- Utilized cmake library dependencies handling to avoid duplicating of `target_link_libraries` calls
- dllexport everything from `cpp_interface.h` and compile it into `bigartm.dll` shared library
- dllexport protobuf messages and compile them into `bigartm.dll` shared library
- Made install target with cmake for both platforms, install cmake config file so that library consumers can `find_package` it without writing special modules
- Disable mandatory install of python package during `make install`

I managed to leave gflags and glog sources without any modifications and integrate them into bigartm build process purely with cmake. Ideally, these should be handled via git submodules.

The only remaining problem is with `Helpers::SetThreadName` function, because `__try {} __except {}` syntax is not available on GCC. I'll look into it later.

Please review this PR. I'm open to cooperation and potentional improvement in order to merge this into master.


maksbotan:
I've fixed all compilation issues revealed by CI and sqaushed all intermediary comments in one.

ofrei:
@maksbotan  Thanks! This change looks great -- I'm happy to merge it into master as it is. @JeanPaulShapo what do you think? 

RE ``Helpers::SetThreadName``, feel free to ignore this. Back in early days I've named treads on Windows to simplify debugging via visual studio - but this is not that important, It is OK if your change breaks tread names, and for sure you don't need to implement this for mingw.

Few things to discuss:
- could you please add some instructions on how to build bigartm on mingw? I've just installed mingw myself, but no more than this. What's preferred way of making  ``boost`` work on mingw? Does you change support python API on mingw, or only native C++ code?
- could you help us to automate mingw build on appveyor? if we don't do this I'm sure the build will broken very soon

maksbotan:
Easiest way to build on MinGW is this: 

- Install msys2 from http://www.msys2.org/
- Run MSYS shell from start menu
- Do `pacman -S mingw-w64-x86_64-{gcc,cmake,boost}
- Run Mingw 64-bit shell from start menu
- `cd` to build dir (msys2 has windows root at `/c/`), run cmake with `-G "MinGW Makefiles"
- Build with `mingw32-make`

I'll try to see if I can do something similar on appveyor.

ofrei:
@maksbotan Thanks! It should be doable with appveyor, they even have mingw pre-installed. There is also a convenient appveyor feature which lets you RDP to their machine and use it for one hour -- one I learned how to do things manually in their environment it is piece of cake to automate them in .appveyor.yml file 
```
# Enable RDP connection to test server as described here: http://www.appveyor.com/docs/how-to/rdp-to-build-worker
on_finish:
- ps: $blockRdp = $true; iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
```

ofrei:
@maksbotan Btw, so far I had back luck with mingw because it didn't have ``mingw32-make.exe``. I've installed ``msys2-x86_64-20161025.exe`` from http://www.msys2.org/ , is this correct or should I use the i686 (e.g. 32 bit version)? Is there a ``pacman`` command to install ``mingw32-make.exe``?

maksbotan:
I hoped it'd be pulled by dependency from cmake. Try `pacman -S mingw-w64-x86_64-make`
Basically, `mingw-w64-` prefix means something native for mingw (as opposed to msys2, which is actually a cygwin - posix port on windows), while `x86_64` means 64-bit variant.
I've not tested compiling it on 32-bit platform and I suppose it's not very relevant anyway :)

ofrei:
My mingw build succeeded (except what depends on python which I didn't have in mingw - but I'll skip testing it for now). I've also run PR build agains our installation scripts, all look green (https://github.com/bigartm/bigartm-install/pull/8)

ofrei:
@maksbotan Merged. I've also noticed you included cpp_interface.cc in the library, and took it out from artm-tests and bigartm cli. This is exactly as things should be, thanks!
It woudl be really great if you have time to look into appveyor build in a separate pull request. Having glog and gtest as submodules is also a great idea :)

PR #858, opened 2017-12-15
AppVeyor config for mingw

maksbotan:
Hi. I finally got time to do this :)

I decided it'd be cleaner to create separate appveor.yml file for mingw: `appveyor-mingw.yml`. You'll need to create new configuration and specify this file. See the results here:
https://ci.appveyor.com/project/maksbotan/bigartm/build/16

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -23,6 +23,13 @@ set(gflags_NAMESPACE "gflags")
 set(gflags_FOUND "ON")
 get_target_property(gflags_INCLUDE_DIR gflags INTERFACE_INCLUDE_DIRECTORIES)
 get_target_property(gflags_LIBRARIES gflags INTERFACE_LINK_LIBRARIES)
+if (NOT gflags_LIBRARIES)
+    # INTERFACE_LINK_LIBRARIES can legitimately be empty, but cmake sets them
+    # as NOTFOUND
+    # This happens e.g. on Alpine Linux
+    set(gflags_LIBRARIES "")
+endif ()
+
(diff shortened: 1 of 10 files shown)


maksbotan:
We have one more thing to decide. Right now compiling generates a lot of warnings like this:

`warning C4275: non dll-interface class 'google::protobuf::Message' used as base for dll-interface class 'artm::TopTokensScore'`

This happens because we are using static libprofobuf.a, but dllexport the generated types (this is required so that consumers can link to our C++ interface). There are two ways to fix this:

- Export protobuf symbols as well (I believe I can do this easily)
- Suppress the warning with `cl.exe` option

I'd prefer the first way, because sometimes consumers will need to call some message's method that is defined in the base class.

ofrei:
Thanks! I'll create new appveyor configuration (after merging the PR). Great to see all info about tests in Appveyor build!

As for warnings, I agree - it sounds important to fit. Protobuf base classes have LIBPROTOBUF_EXPORT definition, but I'm not sure if this will be enough or if there is something else to make it work.

maksbotan:
I'll try to fix it in the next few days. Please don't merge the PR yet.

ofrei:
@maksbotan this slipped out of my hands, but now is a good time to finish :) I can fix warnings myself. Unless you have some further comments I'll probably merge this tomorrow through my rebased PR (https://github.com/bigartm/bigartm/pull/868).

Thanks again for your contribution! CI for mingw and mingw support itself is very very nice to have in place.

maksbotan:
Hi!
Unfortunately, I also didn't have the time to finish this.

As far as I remember, there remain some errors while compiling with MinGW (right now AppVeyour reports failure for this PR). Please give me some time to fix this.

ofrei:
Fixed long time ago

PR #881, opened 2018-02-27
Better support for hierarchical models

ofrei:


ofrei:
@MelLain does this PR look OK to you? 
The idea is to slowly move away from hARTM class because it is too obscure, fragile, and violates `don't multiply entities beyond necessity` principle. Under the hood it creates one ARTM class for each level of hierarchy -- and this is what user want's anyway, e.i. to have each level as ARTM model. Then hARTM offers a lot of strange functionality, like stacking phi and theta matrices across levels, and making sure that different levels have consistent set of parameters. In my opinion users do not benefit from these features. As a downside, hARTM is a place where we could add more bugs, and it makes it harder to save/load models.

I moved core feature of hARTM down to C++ (that is, to create a pseudo-batch of documents from parent level nwt matrix, and add it to fitt_offline). Usage is like this:

```
        # parent level of hierarchy
        level1 = artm.ARTM(num_topics=10, num_document_passes=20,
                           theta_columns_naming='title', seed=100)
        level1.initialize(dictionary=dictionary)
        level1.fit_offline(num_collection_passes=30, batch_vectorizer=batch_vectorizer)

        # child level of hierarchy
        level2 = artm.ARTM(num_topics=70, parent_model=level1, parent_model_weight=0.1,
                           theta_columns_naming='title', seed=101)
        level2.initialize(dictionary=dictionary)
        level2.regularizers.add(artm.HierarchySparsingThetaRegularizer(name='HierSp',
                                                                       tau=-1000.0))
        level2.fit_offline(num_collection_passes=30, batch_vectorizer=batch_vectorizer)
        psi = level2.get_parent_psi()
```

PR #892, opened 2018-03-15
Tool to automatically convert plaintext into UCI Bag-of-words

albertaparicio:
I have been using BigARTM for a couple of days now, and one of the greatest dificulties I have had with it is that there is no easy way to convert a piece of plain text into Bag-of-words or Vowpal Wabbit.

Since I needed it, I made this little tool, which converts any plaintext into BOW. I am sharing it with the community hoping that someone will find it useful.

It can be run from the command-line. It requires a filename of a file with plaintext in it, and it outputs the `docfile.*` and `vocab.*` files right next to the input file.

Cheers!

### DIFF
--- a/utils/text_to_bow.py
+++ b/utils/text_to_bow.py
@@ -6,8 +6,10 @@ from __future__ import print_function
 
 import argparse
 import os
+import re
+from collections import Counter
 
-from sklearn.feature_extraction.text import CountVectorizer
+from six import iteritems
 
 
 def main(args):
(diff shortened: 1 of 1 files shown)


ofrei:
@albertaparicio Thanks a lot! 
This is totally true, we lack good tools to convert raw text into bigartm format. Main reason is because doing converting text properly require a lot of additional processing, like cleaning junk and non-text info, stemming, and  ideally also n-gramm detection. And this is language-specific. But in either case its best to have a basic tool, rather than having nothing. Thanks!

PR #914, opened 2018-06-03
Update T-ARTM

MelLain:


src/artm/messages.proto
```diff
@@ -27,16 +27,16 @@ message Item {
...
   optional string title = 3;
-  repeated int32 token_id = 4;  // obsolete
+  repeated int32 token_id = 4;
   repeated float token_weight = 5;
 
-  // new alternative to token_id field, containa ids of transaction tokens
-  repeated int32 transaction_token_id = 6;
+  // Field contains start index of each transaction from token_id.
+  // In case of regular text will be 0, 1, 2, 3 ... etc.
+  repeated int32 transaction_start_index = 6;
```
ofrei: transaction_start_index - it is convenient to use this variable in for loops like from transaction_start_index(i) to transaction_start_index(i+1) where i is index of transaction. This imply that the length of transaction_start_index should be "number of transactions plus one", and the last element should be "number of tokens". We use this pattern in CsrMatrix.

test_data/vw_transaction_data.txt
```diff
@@ -1,2 +1,2 @@
-doc1 hello world:2 |user^action mel-lain^click:3 mel-lain^show
-doc2 |author mel-lain |@default_class^action^qualifier hello^show^twice:5 world^click^first |@default_class world
+doc1 hello world:2 ||trans1 |user mel-lain:3 |action click:3 || |user mel-lain |action show ||
```
ofrei: Here I don't follow new VW syntax.
* ``||trans1`` indicate that what follows belongs to transaction type ``trans1``  (this identifier can be used to set weight and/or skip certain transaction types from the model). 
* ``||`` indicate start of new transaction. Does it belong to the previously defined transaction type? (e.i. in case of doc1 ``trans1``). 
* Why use ``|`` and not ``@`` in ``|user`` ``|action``? To me these just define class ids.

ofrei: Sorry, it's getting late on a clock... I forgot VW syntax. Surely, we use | to indicate class id.

ofrei: But few more picky questions
- what if first || doesn't indicate transaction type? 
- what if user doesn't close last transaction with trailing ``||`` ?
- does |class scope apply to the next transaction? E.i., in `` |class1 ||trans1 token `` does ``|class1`` apply to token? 

To avoid confusion I suggest to use separate syntax to
- indicate start and/or end of transactions
- indicate scope which defines transaction types

Maybe a good option is to put each transaction like this:
```
doc1 hello world:2 <trans1>|user mel-lain:3 |action click:3</trans1><trans1>|user mel-lain |action show</trans1>
```
or shortcut to 
```
doc1 hello world:2 <trans1 |user mel-lain:3 |action click:3><trans1 |user mel-lain |action show>
```
Perhaps ok to require ``trans1`` in each transaction?

ofrei:
processor_transaction_helpers.cc looks ok. I didn't go into details but it looks good - keep old code as it is (processor_helpers.cc) and handle real transactions in a separate code. The code in processor_transaction_helpers.cc looks ok -- nothing we can't fix later!

src/artm_tests/regularizers_test.cc
```diff
@@ -305,8 +305,8 @@ TEST(Regularizers, NetPlsa) {
 
 // artm_tests.exe --gtest_filter=Regularizers.RelativeRegularization
 TEST(Regularizers, RelativeRegularization) {
-  int nTopics = 500;
-  int nTokens = 500;
+  int nTopics = 50;
```
ofrei: did the test became too slow? (just need to convince myself that introduction of transactions doesn't change results of existing tests)

src/artm/core/phi_matrix_operations.h
```diff
@@ -15,33 +15,7 @@
 namespace artm {
 namespace core {
 
-struct NormalizerKey {
```
ofrei: NormalizerKey is gone.. Good! :)

src/artm/core/token.h
```diff
@@ -10,46 +10,32 @@
...
 
 typedef std::string ClassId;
-const std::string DefaultClass = "@default_class";
-const std::string DocumentsClass = "@documents_class";
+typedef std::string TransactionTypeName;
 
-// Token is a triple of keyword, its class_id (also known as tokens' modality) and type of the transaction.
+const ClassId DefaultClass = "@default_class";
+const ClassId DocumentsClass = "@documents_class";
+const TransactionTypeName DefaultTransactionTypeName = "@default_transaction";
```
ofrei: RE DefaultTransactionTypeName
I think it was a mistake to have ``@default_class`` as a label for default modality. This is quite long and not very useful. 
Is there anything that will go wrong is we set ``DefaultTransactionTypeName`` to an empty string? And eventually fix default class id to "" (and add upgrader to all batches so that @default_class gets changed to "").

ofrei:
:+1:

PR #920, opened 2018-07-29
Fix bulding issues for static CLI

JeanPaulShapo:
This PR is supposed to solve several problems related to building static CLI on Linux:
* Remove autodetection of `libunwind`, because currently there is no static version of this library in official Fedora repositories.
* Enable CMake Policy 0060 (CMP0060) to solve issues with links to  BOOST_IMPORTED_TARGETS
* Move to building shared version of CLI by default.

@ofrei Right now I have little expertise in building it on Windows (e.g. some dll exports, external symbols etc.), and our Appveyor setup can miss some corner cases. Could you help me to ensure that there is no problem with this PR on Windows?
@MichaelSolotky I will be grateful you to check whether this PR works on your Linux machine

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ env:
 before_install:
   # we need latest pip to work with only-binary option
   - pip install -U pip
-  - pip install -U pytest pep8 wheel
+  - pip install -U pytest pep8 wheel pytest-forked
   - pip install -U numpy scipy pandas tqdm --only-binary numpy,scipy,pandas
   - pip install -U protobuf
   # configure ccache
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,18 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.3)
+
+# CMake Policy 0060 (CMP0060) controls how CMake
+# adds library dependencies to linker
+# When this policy is disabled (turned to OLD),
+# CMake uses weird linker options (e.g. -Wl,-Bstatic -lfoo -Wl,-dynamic),
+# which can mess up linkage types of libraries
+# When this policy is enabled (turned to NEW),
+# CMake uses full path when linking libraries,
+# and this helps to avoid aforementioned linker options
+# The complete description of this policy can be found here:
+# https://cmake.org/cmake/help/latest/policy/CMP0060.html
+IF (POLICY CMP0060)
+  cmake_policy(SET CMP0060 NEW)
+endif (POLICY CMP0060)
 
 project(BigARTM)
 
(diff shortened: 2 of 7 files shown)


MichaelSolotky:
Ok, I will check it on arch linux and ubuntu 18.04

ofrei:
@JeanPaulShapo Thanks! This looks good to me. 
- Requiring CMake 3.3 is ok. 
- As long as it builds on Windows appveyor, including bigartm CLI, it's ok. Users on windows should grab existing binary (soon or later via pip install), and not to build themselves. The group of users who actualy want to build on windows is BigARTM developers and contributors. For this group we should improve appveyor build if it doesn't cover some important scenario.
- I still haven't enabled mingw appveyr builds ( https://github.com/bigartm/bigartm/blob/master/appveyor-mingw.yml ).  I'll try to enable them asap because PR like this must be tested against mingw. 
- Disabling forked py.test is ok for windows, but is there a way to avoid if-else switch? Ideally we need "py.test --forked=[YES|NO]", and pass some variable (true on linux, false on windows). If py.test doesn't have this, would it be good idea to (1) make pull requets to py.test , or (2) make our own caller around py.test which adds this small feature? This is a matter of taste, but I don't like duplication in add_test in CMakeLists.txt

ofrei:
I've setup webhooks to trigger both windows builds:
https://ci.appveyor.com/project/bigartm/bigartm-542mi - mingw build
https://ci.appveyor.com/project/bigartm/bigartm - standard windows build (visual studio)
Does these links work for you? 
Also I force push commit to my old PR:
https://github.com/bigartm/bigartm/pull/890
For some reason GitHub shows only one build indicator for appveyor (but both builds are queued at appveyor website). I think github is confused when given repository is linked to two appveyor projects...

JeanPaulShapo:
@ofrei 
* Yes, both links to Appveyor work for me.
I also noted that there were two builds per push on Appveyor for this PR (one for `appveyor.yml` config, and other for `appveyor-mingw.yml`).
* As for `pytest-forked` plugin, I don't see it as permanent solution: it is unmantained and definitely doesn't work on Windows (i.e. you can install it and you `--forked` option, but it simply breaks everything). There are some ways to do it in nicer way (either by adding custom wrapper around `py.test` or by writing some plugins/hooks for py.test), but it can take some time to research. Honestly, I prefer not to modify this solution in this PR, as it's supposed to work and relieves me from Travis CI build issues. But this opinion is rather biased :)

ofrei:
Agree, just keep current way of handling py.test --forked, that's ok.
For mingw, my PR fails with ``mingw-w64-x86_64-gcc-objc: installing mingw-w64-x86_64-gcc (8.2.0-1) breaks dependency 'mingw-w64-x86_64-gcc=7.3.0-2' Command exited with code 1``. I think it was a problem for  a while - so you may ignore it for now. Let's fix appveyour mingw build separately.

JeanPaulShapo:
@ofrei Ok, I see

PR #928, opened 2018-10-17
Fix _run_from_ipython() to differentiate consoles and notebooks

AVBelyy:
Currently, `_run_from_ipython` function in artm_model.py is used to choose between graphical or console version of tqdm progress bars. It makes an erroneous assumption that all IPython interpreters are IPython **notebook** interpreters, however an IPython can be run in non-notebook mode with console CLI, which is **unable** to display graphical progress bars.

This PR aims to fix that. The difference in behavior of new `_run_from_notebook` function is highlighted in the following screenshots.

**CPython:**
screenshot 2018-10-17 at 17 57 35

**IPython console** (previously returned `True`)**:**
screenshot 2018-10-17 at 17 58 01

**IPython notebook:**
screenshot 2018-10-17 at 17 56 47

### DIFF
--- a/python/artm/artm_model.py
+++ b/python/artm/artm_model.py
@@ -51,10 +51,10 @@ PARAMETERS_FILENAME_JSON = 'parameters.json'
 PARAMETERS_FILENAME_BIN = 'parameters.bin'
 
 
-def _run_from_ipython():
+def _run_from_notebook():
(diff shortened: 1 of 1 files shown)


ofrei:
@AVBelyy Thanks for fixing this!

Issue #922, opened 2018-08-02
Rename `async` parameter name to `asynchronous` in Python API to support 3.7


JeanPaulShapo:
According to [Python 3.7 Release Notes](https://docs.python.org/3/whatsnew/3.7.html):

```
* async and await are now reserved keywords.
```

We need to get rid of `async` as paramater name in order to support Python 3.7.

MichaelSolotky:
Here's the solution https://github.com/bigartm/bigartm/pull/931

PR #931, opened 2018-10-27
rename 'async' to 'asynchronous' in order to support python 3.7

MichaelSolotky:
This is the solution to this issue https://github.com/bigartm/bigartm/issues/922

* Now the code can be compiled and models can be built with python 3.7;
* Also there was a bug with coherence computation in the file src/artm/score/top_tokens.cc, it's fixed.

@ofrei, I think we should squash commits in master rather then merge all commits from PR.

### DIFF
--- a/python/artm/artm_model.py
+++ b/python/artm/artm_model.py
@@ -92,8 +92,8 @@ def _topic_selection_regularizer_func(self, regularizers):
 
 
 class ArtmThreadPool(object):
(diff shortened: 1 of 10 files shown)


MichaelSolotky:
what's wrong with appveyor?

ofrei:
@MichaelSolotky RE appveyor, this is happening for a long time, so it needs a ticket #933

PR #941, opened 2018-12-05
fix condition of gathering co-occurrences

MichaelSolotky:


MichaelSolotky:
After MelLain's review there are 2 lines of code that changed in appveyor-mingw.yml. This change fixes crashes that happened in AppVeyor-mingw each time being launched. I think it's ok to merge it.

PR #960, opened 2019-03-24
 add store-symeric-cooc flag 

MichaelSolotky:
There should be a flag available from CLI that allows writing co-occurrences in vowpal wabbit format in 2 different ways:
1) to not store the pair (token_b, token_a) if the pair (token_a, token_b) is stored (store-symeric-cooc is true)
2) to store all the pairs (store-symeric-cooc is false)
Now there's no such flag and value use_symetric_cooc is hardcoded in true


ofrei:
@MichaelSolotky I forgot to merge this PR, sorry! Everything looks good. Please ping me or @MelLain if PR like this one hands unmerged for more than a day

PR #965, opened 2019-04-27
Change base class for ArtmException

bt2901:
The accepted practice for user-defined exceptions is to inherit
Exception class, not BaseException.

Currently, the Python code which tries to handle errors fails to catch
any ArtmException.


ofrei:
@bt2901 :+1:

PR #991, opened 2019-12-12
Revive ci

bt2901:
I was able to fix travis builds, but I have no idea what's wrong with appveyor. Removing it for now.


ofrei:
@bt2901 Please add some comment about the following:
```
for compiler in gcc g++; do ln -s $(which ccache) $HOME/$compiler; done && export PATH=$HOME:$PATH
```
Whatever it does I'm fine with it - but please explain :) Great if you could write it as a comment in .travis.yml script itself.

Also, please restore the .appveyor files. I think we should keep the files in the repository, but disable the build from appveyour website. Do you have access to bigartm projects on appveyour?

ofrei:
P.S. it is awesome to see a green travis build!!!!!

bt2901:
To be honest, I'm not sure I completely understand what this command is doing (it was written by someone before me). My best guess would be that it takes a compiler from `ccache` and declares it to be a replacement to both `gcc` and `g++`. It looks like a part of a failed attempt at testing several versions of different compilers.

I do not seem to have access to bigartm projects on appveyour.

Regarding appveyour, for posterity. The following command works for Python 3 build, but fails for Python 2 build:

```
PostBuildEvent:

  setlocal
  "C:\Program Files (x86)\CMake\bin\ctest.exe" --force-new-ctest-process -C Release
  if %errorlevel% neq 0 goto :cmEnd
  :cmEnd
  endlocal & call :cmErrorLevel %errorlevel% & goto :cmDone
  :cmErrorLevel
  exit /b %1
  :cmDone
  if %errorlevel% neq 0 goto :VCEnd
  :VCEnd
```
It fails with `error MSB3073::<this long code snippet> exited with code 8. `

Unfortunately, I'm completely out of my depth here. Does anybody have any ideas?

bt2901:
After some digging, I'm reasonably certain that the origins of `ln -s` loop were related to some esoteric `clang` problem in another project. I removed the offending workaround and the build is still passing.

bt2901:
Got an invitation, but I failed to find any settings related to this project/

Issue #984, opened 2019-09-11
ARTM loses terms in dictionary via 'dump_artm_model' and 'load_artm_model' operations


ilyakhov:
BigARTM version: 0.10.0

The behaviour depends on .vw input file: with files of the lower sizes (tested on the 1st 5000, 10000 lines of 'vw_file') works fine.
With given file (see file: https://drive.google.com/open?id=1ozTGnvt8L24qiR0grFCrK-hca92WWI6p): the initial vocab size **236785** ; after dumping and loading: **only 27070**  (see code and logs).
Code:

```
import artm

batches_folder = './batches'
batch_vectorizer = artm.BatchVectorizer(data_path=vw_file,
                                        data_format="vowpal_wabbit",
                                        target_folder=batches_folder,
                                        batch_size=1000,
                                        gather_dictionary=False)

dictionary = artm.Dictionary()
dictionary.gather(batches_folder)
dictionary.save_text('./dict')

model_artm = artm.ARTM(num_topics=500,
                       cache_theta=False,
                       reuse_theta=False,
                       seed=42,
                       show_progress_bars=True)

model_artm.initialize(dictionary)

model_artm.dump_artm_model(f'model_non_fit')
print(f'actual vocab size: {len(set(model_artm.phi_.axes[0]))}')
    
m = artm.load_artm_model(f'model_non_fit')
print(f'vocab size after loading: {len(set(m.phi_.axes[0]))}')
```

Logs of dumping:

> I0911 18:54:16.866335  6278 master_component.cc:663] InitializeModel() created matrix pwt with 500 topics and 236785 tokens; 
I0911 18:54:16.866803  6278 c_interface.cc:389] Pass InitializeModelArgs: model_name=nwt, dictionary_name=f5cb8ea1-5857-4d63-aee3-fe522431f343, topic_name_size=500, seed=42 to MN4artm4core15MasterComponentEFvRKNS_19InitializeModelArgsEE
I0911 18:54:20.477962  6278 master_component.cc:663] InitializeModel() created matrix nwt with 500 topics and 236785 tokens; 
I0911 18:54:20.478339  6278 c_interface.cc:231] ArtmCopyRequestedMessage is copying 5477 bytes...
I0911 18:54:20.506392  6278 master_component.cc:502] Exporting score tracker to model_0/score_tracker.bin
I0911 18:54:20.506446  6278 master_component.cc:518] Export of score tracker completed, number of score items: 0
[...]

Issue #953, opened 2019-03-08
ARTM dump and load methods don't work correctly


MichaelSolotky:
Operations of damping and then loading a model give another model with incomplete Phi mathix, as a result, the total number of tokens decreases and columns of Phi don't sum into 1.
![photo_2019-03-08_17-30-28](https://user-images.githubusercontent.com/25284863/54034437-e4892000-41c7-11e9-96c7-a913152ead33.jpg)

vseledkin:
It seems like I facing the same problem, as a result transform for new documents produces poor results, any plans to fix it? Since development process is very slow may be someone have insight how to "workaround it" may be through save/load matrices manually or something like that...

PR #1013, opened 2020-06-27
Fix missing terms in `load_artm_model`

menshikh-iv:
Fix https://github.com/bigartm/bigartm/issues/984
Probably fix https://github.com/bigartm/bigartm/issues/953 too

## Problem description

ARTM splitting large phi matrix to several partitions in `dump_artm_model` -> `MasterComponent::ExportModel` (by protobuf size limitation I guess).
In `load_artm_model` -> `MasterComponent::ImportModel`, `DensePhiMatrix` re-created for every partition, for this reason, instead of full phi matrix, artm load only last partition.

## Solution

Just don't re-create `DensePhiMatrix`, init them only once (and "modify" on every new partition)

### DIFF
--- a/src/artm/core/master_component.cc
+++ b/src/artm/core/master_component.cc
@@ -473,9 +473,12 @@ void MasterComponent::ImportModel(const ImportModelArgs& args) {
     }
 
     topic_model.set_name(args.model_name());
-    target = std::make_shared<DensePhiMatrix>(args.model_name(),
-                                              topic_model.topic_name(),
-                                              instance_->config()->min_sparsity_rate());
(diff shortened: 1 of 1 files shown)


ofrei:
@menshikh-iv Thanks!!
@bt2901 @MelLain could you take a quick look before merging?

bt2901:
I'm not very knowledgeable about internal things like that, but it looks good to me. We indeed were writing `topic_model` inside `target` while re-creating `target` at each chunk.

menshikh-iv:
Thanks for fast merge @ofrei

PR #1015, opened 2020-06-29
Fix get_phi* error when matrix are too large

menshikh-iv:
## Issue

In continuation of #1013 

If you have a **large model** (size(phi_matrix) > int32 limit)  and want to retrieve phi matrix from python like `model.get_phi*()`, you'll get en error 

image

## Reason

When `np.ndarray` passed to C functions, matrix converts to number of bytes and pointer (happens in `artm.wrapper.api.artm_api_call`). Without explicit casting to `int64`, number of bytes automatically converted using `int32`, for this reason, we got negative value as "length" in `ArtmCopyRequestImpl` function.

## Solution

Explicitly cast matrix length with `int64`

### DIFF
--- a/python/artm/wrapper/api.py
+++ b/python/artm/wrapper/api.py
@@ -149,7 +149,7 @@ class LibArtm(object):
                     c_args += [len(message_str), message_cstr_p]
 
                 elif issubclass(arg_type, numpy.ndarray):
-                    c_args += [arg_value.nbytes, ctypes.c_char_p(arg_value.ctypes.data)]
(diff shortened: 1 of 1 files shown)


menshikh-iv:
Ping @bt2901 @MelLain @ofrei, can you have a look, please?

menshikh-iv:
Any updates?

ofrei:
@menshikh-iv Sorry for the delay. I've been hiking (summer vacation) without internet for a while.

PR #1019, opened 2020-08-09
Add topic names for coherence scores (fix #1016)

### LINKED ISSUE #1016: Strange output for TopTokensCoherenceScore
I've create 18 topics for my model and used TopTokensCoherenceScore. When I call `artm_model.score_tracker['TopTokensCoherenceScore'].average_coherence` it gives me correct result for each collection pass, but when I call `artm_model.score_tracker['TopTokensCoherenceScore'].coherence` (or `coherence[-1]` for last collection pass) it gives me output in following format:
```
{'topic_1': [1.1617621183395386,
  1.1602195501327515,
  0.3456096351146698,
  0.9662702679634094,
  0.5320848226547241],
 'topic_2': [0.9413328170776367,
  0.3860290050506592,
  1.2057011127471924,
  1.1175670623779297,
  0.37220320105552673],
 'topic_3': [0.28798189759254456,
  1.1940522193908691,
  0.6581138372421265,
  0.42109599709510803,
  1.9474769830703735],
 'topic_4': [0.5959908366203308, 0.7645043134689331, 0.22957539558410645]}
```
I suppose correct format should be looks like this
```
{'topic_1': 1.1617621183395386,
 'topic_2':1.1602195501327515,
 'topic_3':0.3456096351146698,
 'topic_4':0.9662702679634094,
 'topic_5':0.5320848226547241,
 ...
 'topic_18':0.22957539558410645
 }
```
P.S. bigartm version - 0.10.1

JeanPaulShapo:
Currently there is inappropriate use of `topic_name` field for labeling coherence scores.
While `coherence` field of `TopTokensScore` has one number per requested topics, `topic_name` field has at most `num_tokens` numbers per requested topics.

We may true to use some parts of `topic_name` filed, e.g. we can try to get each `num_tokens`-th element, but there are at least two problems:
* It complicates code
* In fact it's wrong: actual number of gathered top tokens per topic can be less than `num_tokens`, because tokens with small probability cannot be gathered.

Instead of this approach, let's simply add another field into protobuf message `TopTokensScore` which contains topic_names and is aligned with coherence values.

### DIFF
--- a/src/artm/messages.proto
+++ b/src/artm/messages.proto
@@ -360,8 +360,9 @@ message TopTokensScore {
   repeated int32 topic_index = 3;
   repeated string token = 4;
   repeated float weight = 5;
-  repeated float coherence = 6;
-  optional float average_coherence = 7;
+  repeated string coherence_topic_name = 6;
+  repeated float coherence = 7;
+  optional float average_coherence = 8;
 }
 
 // Represents a configuration of a theta snippet score
--- a/python/artm/score_tracker.py
+++ b/python/artm/score_tracker.py
@@ -190,7 +190,9 @@ _set_properties(TopTokensScoreTracker, {'num_tokens': {'proto_name': 'num_entrie
                                                    'proto_qualifier': 'repeated'},
                                         'weights': {'proto_name': 'weight',
                                                     'proto_qualifier': 'repeated'},
-                                        'coherence': {'proto_type': 'array'},
+                                        'coherence': {'proto_type': 'array',
+                                                      'key_field_name': 'coherence_topic_name'},
+
                                         'average_coherence': {}})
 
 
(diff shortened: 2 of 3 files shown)


src/artm/messages.proto
```diff
@@ -360,8 +360,9 @@ message TopTokensScore {
   repeated int32 topic_index = 3;
   repeated string token = 4;
   repeated float weight = 5;
-  repeated float coherence = 6;
-  optional float average_coherence = 7;
+  repeated string coherence_topic_name = 6;
```
JeanPaulShapo, line 363: @ofrei Is it OK to add some backward incompatibility to `TopTokensScore` protobuf message?
Or is it better to just add new field to the end of message?

ofrei:
@JeanPaulShapo I think that's OK in this particular case - can't think of cases where serializing TopTokensScore is critical. I though it's  good to be careful about protobuf serialization (particularly in messages related to Batches, Dictionaries and MasterModelConfig - TopTokensScore is none of those).

